PR #51, opened 2022-12-17
Attempt to locally fix broken Polish songs from usdb.

bohning:


src/usdb_syncer/notes_parser.py
```diff
@@ -409,3 +425,24 @@ def minimum_song_length(self) -> str:
...
+        .replace("³", "ł")
+        .replace("£", "Ł")
+        .replace("ñ", "ń")
+        .replace("Ñ", "Ń")
+        .replace("œ", "ś")
+        .replace("Œ", "Ś")
+        .replace("Ÿ", "ź")
+        .replace("Ź", "Ź")
+        .replace("¿", "ż")
+        .replace("¯", "Ż")
```
RumovZ, line 447: Maybe this could go into a constant we iterate over:
```py
POLISH_REPLACEMENTS = (("¥", "Ą"), ("æ", "ć"), ...)
```
and then
```py
for old, new in POLISH_REPLACEMENTS:
    string = string.replace(old, new)
```
This should be easier to extend if we add more language fixes later.

src/usdb_syncer/notes_parser.py
```diff
@@ -401,6 +401,22 @@ def sanitize(self) -> None:
         self.headers.reset_file_location_headers()
         self.notes.maybe_split_duet_notes()
         self.restore_missing_headers()
+        if self.headers.language == "Polish":
+            self.fix_polish_song()
+
+    def fix_polish_song(self) -> None:
+        self.headers.artist = _fix_polish_letters(self.headers.artist)
+        self.headers.title = _fix_polish_letters(self.headers.title)
+        # edition, genre, creator, p1, p2, album, comment, anything else?
```
RumovZ, line 410: I guess we could just iterate over all headers indiscriminately, but I'd have to investigate how that would work myself. Unless you want to do it yourself, I'll have a look later.

RumovZ (commented):
Looks great overall!

src/usdb_syncer/notes_parser.py
bohning: I tried to iterate over the respective headers with something like `for tag in [self.headers.artist, self.headers.title, ...]: tag = _fix_polish_letters(tag)`, but that didn’t work.

PR #57, opened 2022-12-19
Implement fix_low_bpm(), fix_line_breaks() and fix_touching_notes().

bohning:
fix_low_bpm(): (repeatedly) doubles BPM and all note timings until it is equal to or above a certain threshold (currently 200.0).
fix_line_breaks(): sets sensible line break timings (similar to the timings correction in USDX editor via "T")
fix_touching_notes(): shortens the earlier of two notes by 1 beat if it touches the next note (and it’s duration is longer than 1).


src/usdb_syncer/notes_parser.py
```diff
@@ -438,20 +453,104 @@ def minimum_song_length(self) -> str:
...
+        until the BPM is in a range (200, 400)"""
+        bpm_threshold = 200.0
+
+        if (original_bpm := self.headers.bpm) >= bpm_threshold:
+            return
+
+        while self.headers.bpm < bpm_threshold:
+            self.headers.bpm *= 2
+
+        timing_factor = int(self.headers.bpm / original_bpm)
```
RumovZ, line 481: Maybe more elegantly:
```py
import math

# how often to multiply bpm until it is larger or equal to the threshold
exp = math.ceil(math.log2(bpm_threshold / self.headers.bpm))
factor = 2**exp
```

src/usdb_syncer/notes_parser.py
```diff
@@ -438,20 +453,104 @@ def minimum_song_length(self) -> str:
...
+        # modify all note timings
+        for line in self.notes.all_lines():
+            line.multiply(timing_factor)
+
+    def fix_line_breaks(self) -> None:
+        for i, line in enumerate(self.notes.player_1):
+            if line.line_break:
+                # remove end
+                line.line_break.end = None
+                # set line break timing similar to USDX implementation
```
RumovZ, line 499: Could you add a link to that part in USDX's source code?

src/usdb_syncer/notes_parser.py
```diff
@@ -438,20 +453,104 @@ def minimum_song_length(self) -> str:
...
+                    # set line break timing similar to USDX implementation
+                    next_line_start = self.notes.player_1[i + 1].start()
+                    gap = next_line_start - line.end()
+
+                    if gap < 2:
+                        line.line_break.start = next_line_start
+                    elif gap == 2:
+                        line.line_break.start = line.end() + 1
+                    else:
+                        line.line_break.start = line.end() + 2
```
RumovZ, line 524: You can extract the for loop into a function and call it for both player_1 and player_2 to avoid the duplication.

If you want to use the `all_lines()` iterator, one option would be to store a refefernce to the last line:
```py
last_line = None
for line in self.notes.all_lines():
    if not (last_line and last_line.line_break):
        continue
    # fix break of last_line
    ...
    last_line = line
```

src/usdb_syncer/notes_parser.py
```diff
@@ -438,20 +453,104 @@ def minimum_song_length(self) -> str:
...
+        for i, line in enumerate(self.notes.player_1):
+            for j, note in enumerate(line.notes[:-1]):
+                if note.end() == line.notes[j + 1].start and note.duration > 1:
+                    note.duration -= 1
+            if not line.is_last():
+                if (
+                    line.end() == self.notes.player_1[i + 1].start
+                    and line.notes[-1].duration > 1
+                ):
+                    line.notes[-1].duration -= 1
```
RumovZ, line 536: Maybe we could use a method on `Note` like this:
```py
def shorten(self):
    if self.duration > 1:
        self.duration -= 1
```

src/usdb_syncer/notes_parser.py
```diff
@@ -438,20 +453,104 @@ def minimum_song_length(self) -> str:
...
+            for i, line in enumerate(self.notes.player_2):
+                for j, note in enumerate(line.notes[:-1]):
+                    if note.end() == line.notes[j + 1].start and note.duration > 1:
+                        note.duration -= 1
+                if not line.is_last():
+                    if (
+                        line.end() == self.notes.player_1[i + 1].start
+                        and line.notes[-1].duration > 1
+                    ):
+                        line.notes[-1].duration -= 1
```
RumovZ, line 548: Same as above: iterate all lines together or extract a function.

tests/resources/txt/fixes/low_bpm_touching_notes.out.txt
```diff
@@ -0,0 +1,193 @@
...
+: 2372 16 12 ter
+- 2390
+: 2428 7 24 And
+: 2436 7 26  you'll
+: 2444 11 28  sure
+: 2456 3 24 ly
+: 2460 24 21  find
+: 2492 15 24  a
+: 2508 64 24  way
+E
```
RumovZ, line 193: Could you reduce the txts to the amount required for the tests to work?

RumovZ (changes requested):
Great work!
I think you appreciate the feedback so I'm being nitpicky. :)

src/usdb_syncer/notes_parser.py
```diff
@@ -438,20 +453,104 @@ def minimum_song_length(self) -> str:
...
         """
         if (offset := self.notes.start()) == 0:
+            # round GAP to nearest 10 ms
+            self.headers.gap = round(self.headers.gap / 10) * 10
             return
         for line in self.notes.all_lines():
             line.shift(-offset)
-        self.headers.gap = self.headers.gap + beats_to_ms(offset, self.headers.bpm)
+        self.headers.gap = (
+            round((self.headers.gap + beats_to_ms(offset, self.headers.bpm)) / 10) * 10
```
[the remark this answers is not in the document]
bohning: Ah, nice!

src/usdb_syncer/notes_parser.py
```diff
@@ -438,20 +453,104 @@ def minimum_song_length(self) -> str:
...
+        )
+
+    def fix_low_bpm(self) -> None:
+        """(repeatedly) multiplies BPM value and all note timings by two
+        until the BPM is in a range (200, 400)"""
+        bpm_threshold = 200.0
```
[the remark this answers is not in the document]
bohning: "make it a constant" means just capitalizing it? And where do I move the constant? All the way to the top of the file, after the imports?

src/usdb_syncer/notes_parser.py
bohning: How to do that with the line limit length = 88?

https://github.com/UltraStar-Deluxe/USDX/blob/0974aadaa747a5ce7f1f094908e669209641b5d4/src/screens/UScreenEditSub.pas#L2976

src/usdb_syncer/notes_parser.py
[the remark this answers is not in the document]
RumovZ: Yes, exactly.

src/usdb_syncer/notes_parser.py
RumovZ: You can add ` # pylint: disable=line-too-long` to the end of the line.

src/usdb_syncer/notes_parser.py
bohning: In fact I had that in mind, but then was unsure if I should import math for that when a simple while-loop does the job =) It is now using math’s ceil() and log2() expression.

src/usdb_syncer/notes_parser.py
```diff
@@ -517,6 +523,18 @@ def fix_touching_notes(self) -> None:
                     if line.end() == self.notes.track_1[num_line + 1].start:
                         line.notes[-1].shorten()
 
+    def fix_pitch_values(self) -> None:
+        min_pitch = 1000
+        for note in self.notes.all_notes():
+            if min_pitch is None or note.pitch < min_pitch:
+                min_pitch = note.pitch
```
RumovZ, line 530: I believe `min_pitch = min(note.pitch for note in self.notes.all_notes())` should also work.

src/usdb_syncer/notes_parser.py
```diff
@@ -456,6 +461,7 @@ def fix(self) -> None:
         self.fix_low_bpm()
         self.fix_line_breaks()
         self.fix_touching_notes()
+        self.fix_pitch_values()
 
     def minimum_song_length(self) -> str:
```
RumovZ, line 466: I think it's considered best practice not to pass in more than the minimum required arguments into a function. In the case of a method, this would mean passing in the 'smallest self'.
In more understandable terms: You can move this method into `Tracks` since `self` is only ever used to refer to `self.notes`.

RumovZ (approved):
🚀

src/usdb_syncer/notes_parser.py
bohning: As always, thanks for the valuable feedback!

PR #58, opened 2022-12-19
Update README.md

mjhalwa:
adds instructions for setup in windows using wsl 2

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,30 +27,8 @@ The package has a defined entry point for the GUI. Simply type in `usdb_syncer` 
 
 On __Windows__, this might not work out of the box, because symlinks are disabled by default.
 Try either
-1. following this answer on Stack Overflow: <https://stackoverflow.com/a/59761201/7542749> or
-2. setup with WSL 2 and e.g. Ubuntu 22.04
-
-    [latest WSL updates support GUIs started from WSL](https://learn.microsoft.com/en-us/windows/wsl/tutorials/gui-apps). Make sure to to use WSL 2 and update to latest version. On Windows run
-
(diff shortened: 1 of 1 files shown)


RumovZ:
Cool, thanks for figuring this out and sharing!
I think even if you have developer mode enabled and symlinks working (or if we remove the symlink in the future), this is a very useful guide, e.g. if you want to debug a UNIX error on a Windows machine. Maybe we can create a separate wiki page for it? Then in the README, we could link to that page.

mjhalwa:
If you asked me:

Wiki or readme, both is fine for me, only I think i am not able to propose a wiki change via PR (can i?) in the main repo. So if you add an article in the wiki, I could modify my proposed change to the readme. Maybe you add a wiki page for "Setup" or "Installation" with chapters for each OS. I guess my originally proposed change to the readme would fit a "Windows" chapter.

If I understood correctly, I can simply add another commit to the feature branch in my repo and it is added to this PR automatically.

I'd change the readme to something like

``` md
either
- enable developer settings and use git's symlinks capabilities, see [this stackoverflow answer](https://example.com) or
- use WSL with at least Ubuntu 22.04, see [detailed instructions in wiki](https://example.com)
```

Where the wiki page link points to the "Windows" Chapter on a wiki page

RumovZ:
Yes, exactly. Maybe @bohning can enable wiki edits for non-contributors?

bohning:
> Yes, exactly. Maybe @bohning can enable wiki edits for non-contributors?

Just enabled the Wiki for non-contributors and also added @mjhalwa as a contributor.

PR #60, opened 2022-12-23
Fix emojis breaking plain URLs.

### LINKED ISSUE #59: Fix "broken" plain urls in usdb comments
Usdb replaces certain character combinations in comments by emoticons. Some of these can appear in URLs, e.g. =P and XD/xD. The Syncer should handle these correctly, substituting the emoticon character with the corresponding original characters.

bohning:
Emojis are replaced with their character counterparts in order to restore functional URLs. Fixes #59

### DIFF
--- a/src/usdb_syncer/usdb_scraper.py
+++ b/src/usdb_syncer/usdb_scraper.py
@@ -350,7 +350,12 @@ def _parse_comments_table(comments_table: BeautifulSoup) -> list[SongComment]:
 
 
 def _parse_comment_contents(contents: BeautifulSoup) -> CommentContents:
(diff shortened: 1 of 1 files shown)


RumovZ (approved):
Wow, what a simple solution!
And we get a test for free. :-)

bohning:
Yeah, looks like a three-minute fix, but actually took me much longer find out, debug and finalize. I forgot to delete the commented out line (previous code).

RumovZ:
> Yeah, looks like a three-minute fix, but actually took me much longer find out, debug and finalize.

Isn't it always like this? 😅 

> I forgot to delete the commented out line (previous code).

Oh right. I saw it and forgot about it, too.

PR #62, opened 2022-12-23
Gracefully handle empty notes (no or whitespace text)

bohning:
Unless the note is a freestyle note, use ~ for empty notes.

### DIFF
--- a/src/usdb_syncer/notes_parser.py
+++ b/src/usdb_syncer/notes_parser.py
@@ -54,6 +54,8 @@ class Note:
         if kind != NoteKind.FREESTYLE:
             if not text.strip():
                 text = "~" + text
+        if text.strip() == "-":
+            text = text.replace("-", "~")
         return Note(kind, start, duration, pitch, text)
 
     def __str__(self) -> str:
--- a/src/usdb_syncer/notes_parser.py
+++ b/src/usdb_syncer/notes_parser.py
@@ -40,7 +40,7 @@ class Note:
 
     @classmethod
     def parse(cls, value: str) -> Note:
-        regex = re.compile(r"(:|\*|F|R|G):? +(-?\d+) +(\d+) +(-?\d+) (.+)")
+        regex = re.compile(r"(:|\*|F|R|G):? +(-?\d+) +(\d+) +(-?\d+) (.*)")
         if not (match := regex.fullmatch(value)):
             raise NotesParseError(f"invalid note: '{value}'")
         text = match.group(5)
@@ -51,6 +51,9 @@ class Note:
             pitch = int(match.group(4))
         except ValueError as err:
             raise NotesParseError(f"invalid note: '{value}'") from err
+        if kind != NoteKind.FREESTYLE:
+            if not text.strip():
+                text = "~" + text
         return Note(kind, start, duration, pitch, text)
 
     def __str__(self) -> str:
--- a/src/usdb_syncer/notes_parser.py
+++ b/src/usdb_syncer/notes_parser.py
@@ -40,10 +40,12 @@ class Note:
 
     @classmethod
     def parse(cls, value: str) -> Note:
-        regex = re.compile(r"(:|\*|F|R|G):? +(-?\d+) +(\d+) +(-?\d+) (.*)")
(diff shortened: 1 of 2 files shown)


src/usdb_syncer/notes_parser.py
```diff
@@ -51,6 +51,11 @@ def parse(cls, value: str) -> Note:
             pitch = int(match.group(4))
         except ValueError as err:
             raise NotesParseError(f"invalid note: '{value}'") from err
+        if kind != NoteKind.FREESTYLE:
```
RumovZ, line 54: What about the opposite case? Shouldn't we keep deleting freestyle notes without text?

RumovZ (approved):

src/usdb_syncer/notes_parser.py
bohning: Good point. I guess we should.

bohning: I also just stumbled upon a note of the form `: 1 2 3` (missing space after duration and missing text) which should also be handled as an empty note that needs a `~`.

RumovZ: Yes, as far as I can see, there is no reason to keep empty freestyle notes.

Oh, right. The last space also has to be made optional.

bohning: So then `regex = re.compile(r"(:|\*|F|R|G):? +(-?\d+) +(\d+) +(-?\d+) ?(.*)?")` (also the last capture group optional?)

RumovZ: The last `?` is redundant, because `(.*)` will always match (the empty string).
I would say, if there are lyrics they must be separated by a space. The above regex would match `: 1 1 1foo` for example.
That would mean something like this: `(:|\*|F|R|G):? +(-?\d+) +(\d+) +(-?\d+)(?: (.*))?`. But note that now `match.group(4)` can be `None`.

RumovZ: On a different note, `parse()` was an `__init__()` previously, so I couldn't return `Note | None` and had to raise an exception in case the note could not be parsed. If it makes things easier, you could remove the exception, move the logging to `Note.parse()` and return `None` on a parse failure.

bohning: Yes, I agree that if there are lyrics, they must be separated by a space and we should reject `: 1 1 1foo` as invalid. It should only work for empty notes if there is no space after the pitch (which should then get ~ as text).

Thus:
`: 1 1 1` --> `: 1 1 1 ~`
`: 1 1 1 `--> `: 1 1 1 ~`
`: 1 1 1  ` --> `: 1 1 1 ~ `

bohning:
I tested the current code with http://usdb.animux.de/index.php?link=detail&id=25213 and seems to work flawlessy.

src/usdb_syncer/notes_parser.py
```diff
@@ -40,17 +40,24 @@ class Note:
 
     @classmethod
     def parse(cls, value: str) -> Note:
-        regex = re.compile(r"(:|\*|F|R|G):? +(-?\d+) +(\d+) +(-?\d+) (.+)")
+        regex = re.compile(r"(:|\*|F|R|G):? +(-?\d+) +(\d+) +(-?\d+)(?: (.*))?")
         if not (match := regex.fullmatch(value)):
             raise NotesParseError(f"invalid note: '{value}'")
         text = match.group(5)
+        if not text:
+            text = ""
```
RumovZ, line 48: You can also write `text = match.group(5) or ""`. (I hope pylint understands that as well.)

src/usdb_syncer/notes_parser.py
bohning: Where did your other remarks go?

RumovZ: Somehow the other comment ended up in a different thread. 🤔

bohning: > On a different note, `parse()` was an `__init__()` previously, so I couldn't return `Note | None` and had to raise an exception in case the note could not be parsed. If it makes things easier, you could remove the exception, move the logging to `Note.parse()` and return `None` on a parse failure.

Not sure if I fully understand this. Let’s exchange on this in a call some time.

PR #64, opened 2022-12-26
ci: Create releases with GitHub actions

### LINKED ISSUE #27: Setup automatic build system
Standalone executables for the three major OS.

g3n35i5:
This resolves #27


.github/workflows/release.yaml
```diff
@@ -17,10 +17,10 @@ jobs:
             PYINSTALLER_ARGS: "--onefile"
           - os: macos-latest
             APP_NAME: "USDB Syncer.app"
```
bohning, line 19: APP_NAME should be "USDB Syncer", right? The `.app` is appended as part of the `pyinstaller --windowed` process, I believe.

.vscode/settings.json
```diff
@@ -23,7 +24,5 @@
   },
   "python.languageServer": "Default",
   "python.analysis.typeCheckingMode": "basic",
-  "editor.rulers": [
-    88
-  ]
-}
\ No newline at end of file
+  "editor.rulers": [120]
```
RumovZ, line 27: Was there any particular reason you changed this? It does'nt conform to our configured formatting rules.

RumovZ (changes requested):
Great job, works like a charm on Windows!

Do you know why Windows Defender would suddenly warn about running the app? It didn't with the previous exe.
It would also be convenient if we could skip the archiving on Windows and Linux. But not a big deal.

RumovZ (approved):

g3n35i5:
> Great job, works like a charm on Windows!
> 
> Do you know why Windows Defender would suddenly warn about running the app? It didn't with the previous exe. It would also be convenient if we could skip the archiving on Windows and Linux. But not a big deal.

I would leave this as a ZIP for two reasons: It saves storage space and unifies all platforms. Otherwise you always have to consider special conditions "If on Windows/Linux then don't zip, otherwise do".

On the subject of Windows Defender: I think that the warning should always come, because the exe is not signed?

Issue #21, opened 2022-11-15
Audio normalization


bohning:
One problem I have often encountered is that the loudness is different in each song, so maybe usdb_dl can alleviate the problem by performing some kind of audio normalization on the song collection.

Maybe this could be helpful:

- https://wiki.tnonline.net/w/Blog/Audio_normalization_with_FFmpeg
- https://github.com/slhck/ffmpeg-normalize/
- https://github.com/marcosdotme/auto-2pass-loudnorm
- https://github.com/chaudum/rgain3
- https://github.com/jiaaro/pydub
- https://stackoverflow.com/questions/42492246/how-to-normalize-the-volume-of-an-audio-file-in-python
- https://bbs.archlinux.org/viewtopic.php?id=269966
- https://github.com/Moonbase59/loudgain
- https://github.com/dgilman/aacgain
- https://github.com/slhck/ffmpeg-normalize
- https://github.com/yt-dlp/yt-dlp/issues/3409

mjhalwa:
Fot future debugging/testing (based on the currently siggested youtube-link):

- a __quiet__ song: http://usdb.animux.de/index.php?link=detail&id=3453
- a __variable loudness__ song starting loud, but having several quiet parts as well: http://usdb.animux.de/index.php?link=detail&id=5846
- a __loud__ song: http://usdb.animux.de/index.php?link=detail&id=19086

RumovZ:
Closed in #88.

Issue #37, opened 2022-12-04
Remove relative paths


RumovZ:
Relative paths are bad practice, as they can lead to permission issues, annoy users and possibly crash the app in macOS bundles.
Currently known relative paths:
- Song PDF
- Log file

PR #65, opened 2022-12-29
add audio normalization to fix issue #21

mjhalwa:


src/usdb_syncer/resource_dl.py
```diff
@@ -79,12 +79,23 @@ def download_video(
...
         }
-        ydl_opts["postprocessors"] = [postprocessor]
+        pp_normalize = {
+            "key": "ExecAfterDownload",
+            "exec_cmd": " && ".join(
+                [
+                    # simple loudness normalization with EBU R128 algorithm
+                    # see https://trac.ffmpeg.org/wiki/AudioVolume
+                    f"ffmpeg -i {path_stem}.%(ext)s -filter:a loudnorm temp-audio.%(ext)s",
+                    f"mv test-audio.tmp.%(ext)s {path_stem}.%(ext)s",
```
RumovZ, line 94: Haven't had a chance to try this out, yet, but doesn't `path_stem` have to be escaped? And I think the temp file should be created in the same folder, or else there may be issues when moving across filesystems or crashing bundles as in #37.

g3n35i5: Additionally, we should be careful to use OS-specific commandline arguments as little as possible. I would suggest at this point a python [tempfile ](https://docs.python.org/3/library/tempfile.html)and the [shutil ](https://docs.python.org/3/library/shutil.html)library for file handling :)

g3n35i5 (changes requested):

mjhalwa:
hmm... probably `mv` just worked on my Windows machine because I have WSL installed.

Meanwhile I checked a download in USDX - which seemed to be fine in first place, as all files were there, but USDX did not find the file. I figured out the `#MP3` Header in the txt was missing. Seems I missinterpreted the CLI output.  It also needs quotes (to handle the spaces in filename and path)... I try to fix this. Then I check the `shutil` and `tempfile`

mjhalwa:
Using `shutil` and `tempfile` would require extra python code. As ffmpeg has to be installed anyways (platform-independent) what about using `ffmpeg` instead of `mv`?

- copy file as is

  ```
  ffmpeg -i tmp.m4a -acodec copy -y output.m4a
  ```

  - `-acodec copy` ... saves time and file seems to be almost identical
  - `-y` ... replace `output.m4a` in case it already exists

unfortunately no command for deleting the input file after conversion, but this could be done with `shutils` afterwards

mjhalwa:
... forgot the `finally`

mjhalwa:
I checked the `build error` but I don't see the relation to my changes. Is this really related to the change in this commit?

g3n35i5 (approved):
I am currently unable to test this PR on my system, but from a code review standpoint this looks much better now, great job!

RumovZ:
Thanks for the improvements!

What exceptions do you expect that require a `finally`? Does yt_dlp raise something different than `YoutubeDLError` during postprocessing?

It's not a requirement for this PR, because other code should be using it as well, but using `tempfile` should be possible here with `NamedTemporaryFile(..., delete=False)`.

mjhalwa:
Regarding the `finally`: I do not think of any particular error. I was just thinking of possible flows at that point. In case the second ffmpeg command fails for an unexpected, but exception raising reason, the temporary file created in first place is not deleted on an else, where the `finally` will delete a present temporary file anyways.

Regarding `tempfile`: what i mean with "extra python code" is (but probably I have a misconception): I could use `tempfile`, `os` and/or `shutil` **after** the download process by adding extra python code at the end of the modified function on success, operating on the existing file. But what I tried was to utilize another postprocesser that is handled by the external package `yt_dlp` and is invoked automatically. I don't see a way to add `tempfile` directly to the postprocessor dict in order to avoid the generation of the temporary file.

RumovZ:
Sorry for the delay, I finally tried this out.
Unfortunately, it doesn't seem to do what we want. Instead of adjusting the volume across multiple files, it compresses the audio of single files. We only want to add a constant gain to the whole file, not alter the ration between different parts of it, right?

grafik

In this example, the initial 15 seconds are just background noise. As you can see in the lower track, its volume gets boosted and the actual music part is toned down.

mjhalwa:
I Downloaded [Sonic Visualizer](https://www.sonicvisualiser.org/download.html) and did some own analysis:

4 songs:

1. `blue` [Disney's The Lion King - Circle Of Life](https://www.youtube.com/watch?v=GibiNy4d4gc)
2. `green` [Dragonball Z - Cha la Head Cha la](https://www.youtube.com/watch?v=ulLm6JCVXuQ)
3. `purple` [Die Ärzte - Männer sind Schweine](https://www.youtube.com/watch?v=404oPn6tudE)
4. `orange` [Disney's Der König Der Löwen - Kann es wirklich Liebe sein](https://www.youtube.com/watch?v=c4UWl7sYZP4)

These are the 4 songs in **original state**:

grafik

If you listen to them without adjusting volume: 

1. circle of life has loud and silent parts put is kind of fine
2. dragonball is super loud the first base kind of hurts in the ear
3. "Männer sind Schweine" is too quite (better was... when i downloaded it some weeks ago it was quite, but it seems to be louder now on youtube)
4. is so silent you cannot here the very beginning

These are the 4 songs **after normalization**:

grafik

If i listen to all 4 without adjusting volume they seem about the same loudness (compared to before)

RumovZ:
I'm no expert here, either. 🙂 
The problem is not that ffmpeg is using some clever algorithm to detect the perceived loudness, but that different frames of the same mp3 file receive different gains. Compare that to Mp3Gain, which also uses EBU R128, but assigns a constant gain to the entire file.
We can assume that our files are professionally mixed audio tracks and we don't want to mess with them, we just want to play back different files with about the same (perceived) loudness.
https://github.com/chaudum/rgain3 looks promising for that purpose, but I haven't checked it out in detail.

mjhalwa:
ok I'll try to compare to some other programms and see what the results look and listen like. I reply, if I have news on this. I have not tested other approaches yet 

Still, please compare the actual sound of my four samples (or some own) by loading them with the current syncer and my feature branch and listening to them. I check by listening to the first 10 seconds of the four original files, then listen to the first 10 seconds of the normalized files. I can clearly hear that songs are now at about the same loudness level. Even if checking different parts of the normalized songs, it does not sound like anything is messed up in the song (even though I understand that the analysis looks like) and that ffmpeg does fulfill the goal of avoiding to tab the volume button on every single song I play.

RumovZ:
> it does not sound like anything is messed up in the song

Maybe it doesn't make much of a difference for those songs, I can check tomorrow if you want. But in my example (Die Ärzte - Monsterparty), you can clearly hear that the background noise in the beginning got very loud and the actual music quiet.

> and that ffmpeg does fulfill the goal of avoiding to tab the volume button on every single song I play.

Of course, the issue is that there is *too much* normalization (within files as well as across files) not too _few_.

bohning:
I wanted to look into `rgain3`, but `pip install rgain3` fails with
```
...
Building wheel for PyGObject (pyproject.toml) ... error
...
In file included from gi/gimodule.c:34:
      In file included from /private/var/folders/lx/fkd4dczx16q49q80shrgdvc00000gn/T/pip-install-30gm84n6/pygobject_b3e381345f3b4b42a1dac6dac507f30c/gi/pygi-error.h:26:
      In file included from /private/var/folders/lx/fkd4dczx16q49q80shrgdvc00000gn/T/pip-install-30gm84n6/pygobject_b3e381345f3b4b42a1dac6dac507f30c/gi/pygi-cache.h:26:
      /usr/local/Cellar/gobject-introspection/1.74.0/include/gobject-introspection-1.0/girffi.h:25:10: fatal error: 'ffi.h' file not found
      #include <ffi.h>
               ^~~~~~~
      1 error generated.
      error: command '/usr/bin/clang' failed with exit code 1
```

mjhalwa:
finally linear normalization with 2 passes to ffmpeg

bohning:
Superseded by #88 .

Issue #36, opened 2022-12-02
Sorting songs by views (numeric) sorts in a wrong way


g3n35i5:
When sorting the songs by views, they're getting sorted by the following scheme:

1
11
111
2
22
222

PR #66, opened 2022-12-30
Fix issue 36 sorting for views and local files

BWagener:
This fixes https://github.com/bohning/usdb_syncer/issues/36

Sorting behavior for other columns should remain unchanged.

Please let me know what you think about overriding the `QSortFilterProxyModel`'s `lessThan` method instead of the solution in the initial PR. It would facilitate locale aware sorting which would be especially useful for the title and artist columns.
An example for this is given in the [docs](https://doc.qt.io/qtforpython-5/PySide2/QtCore/QSortFilterProxyModel.html#sorting) at the bottom of the big code block in the section about sorting.


bohning:
I just tested it and works like a charm. I’ll let @RumovZ give the final go for the merge. Thanks a lot for contributing!

RumovZ:
Great, thanks a lot!
Let's just keep an eye on demand for locale-aware sorting for now. One alternative would be to reuse the fuzzy search text for sorting. Not sure if this would be the desired behaviour, just putting it out there.

Issue #42, opened 2022-12-09
On first start, ask user for target karaoke software to determine suitable text encoding, audio and video formats


bohning:
Different karaoke softwares support different text file encodings, audio and video formats. The user should select the target karaoke software so the Syncer downloads suitable encodings/formats.

Related: https://github.com/bohning/usdb_syncer/wiki/Comparison-of-different-Karaoke-software

**Line Ending**
either always CRLF (supported by all karaoke softwares? But may cause issues in Linux/Mac text editors?) or Windows -> CRLF, Mac/Linux -> LF

**Text File Encoding**
USDX < 1.1 / CMD (legacy support) -> CP1252
USDX and variants, Play, Performous -> UTF8
Vocaluxe -> UTF8-BOM

**Audio**
USDX < 1.1 / CMD (legacy support) -> mp3
USDX and variants, Vocaluxe, Performous -> m4a
Play -> mp3

**Video**
USDX < 1.1 / CMD (legacy support) -> TBD
USDX and variants -> TBD
Vocaluxe -> TBD
Performous -> TBD
Play -> TBD (depending on OS? Unity supports different codecs/containers for different OS)

bohning:
We also have to keep in mind that users may want to use other tools that the karaoke software with the downloaded songs (e.g. to improve them), and these tools may support other codecs/encodings. As a concrete example, users may use USDX for playing (supports m4a), but want to use YASS for editing (does not support m4a, supports mp3, maybe ogg?).

PR #68, opened 2022-12-31
First implementation of fix_apostrophes().

bohning:
I didn’t know how to iterate over the headers, so this has a lot of duplicate code in it. Maybe wrong apostrophes in header tags should be replaced directly when they are read?

Pylint complains:
```
src/usdb_syncer/notes_parser.py:407:12: C0103: Attribute name "p1" doesn't conform to snake_case naming style (invalid-name)
src/usdb_syncer/notes_parser.py:409:12: C0103: Attribute name "p2" doesn't conform to snake_case naming style (invalid-name)
```
but I don’t how to deal with that as these are the given header tag name variables.


RumovZ:
You can add `p1` and `p2` to `good-names` in `pyproject.toml`.
Isn't `'` sometimes used as a quotation mark as well?
You could do something like this to avoid the duplication:
```py
for key in ("artist", "title", ...):
    if value := getattr(self, key):
            setattr(self, key, value.replace("`", "’").replace("´", "’").replace("'", "’"))
```
(From memory, didn't check this code.)
Using `getattr()` and `setattr()` is not ideal, since we lose a lot of type safety, but I chose it as the lesser evil here. Maybe I or someone else will come up with a better solution in the future.

bohning:
> Isn't `'` sometimes used as a quotation mark as well?

I can’t rule that out for all songs, but I’ve mostly seen regular (`"`), typographic (`“` and `”`) or french (`»` and `«`) quotation marks. I remember that some early SingStar version used double `'` instead of regular quotation marks. I just pushed a commit for those. I don’t see a way how to differentiate between `'` being used as a quotation mark or as an apostrophe (it can be at the start of the word, e.g. `'cause` or at the end, e.g. `somethin'`, so testing for whitespace before or after the single quotation mark won’t help).

RumovZ:
> I don’t see a way how to differentiate between ' being used as a quotation mark or as an apostrophe

Hm, I guess there isn't. You could check for the single `'word'` case, but that quickly breaks apart with `'more words'`.
I think single quotes are often preferred in English and `’this’` would look even worse than `'this'`, while `‘this’` would be ideal.
If you like, I can try to figure out how common `'` is as a quotation mark in USDB songs in the next couple of days. Hopefully, it's not very common and we can ignore this use case.

bohning:
Yes, let’s do it that way. If it’s not too many, I can manually fix them and we can ignore the use case for new songs.

RumovZ:
Another thing we should be aware of: This fix can introduce non-ASCII characters to currently pure ASCII txts. Especially English song creators usually didn't have to worry about encoding until now and may run into issues when uploading UTF-8 to USDB.

bohning:
Good point. I propose we discuss potential issues in a call. I have been using the typewriter’s apostrophe for years and never had any issues uploading to usdb - neither via clipboard nor via the usdb upload page.

mjhalwa:
> > I don’t see a way how to differentiate between ' being used as a quotation mark or as an apostrophe
> 
> Hm, I guess there isn't. You could check for the single `'word'` case, but that quickly breaks apart with `'more words'`. I think single quotes are often preferred in English and `’this’` would look even worse than `'this'`, while `‘this’` would be ideal. If you like, I can try to figure out how common `'` is as a quotation mark in USDB songs in the next couple of days. Hopefully, it's not very common and we can ignore this use case.

Probably, you have already thought of this, but may we identify apostrophes and quotes by some rule set like:

- treat all appearing forms as either quote or apostrophe
- quotes appear in pairs: 1. at a starting word, 2. at ending of a following word
- only apostrophes are within words
- apostrophes are after words, if there is not currently an unpaired 1. quote 

Problem however: apostrophes within quotes :/

RumovZ:
> apostrophes are after words, if there is not currently an unpaired 1. quote

Apostrophes may also appear at the beginning of words though, e.g. German `eine`→ `’ne`.

bohning:
Or in English `’em`, `’cause`

bohning:
From my experience and gut feeling, `'` used as single upright quotation mark are a lot less common than `'` used as apostrophe.

bohning:
> I remember that some early SingStar version used double `'` instead of regular quotation marks. I just pushed a commit for those.

Just because I stumble upon this just now: use of two single upright quotation marks instead of `"` used in SingStar songs:
image

bohning:
> Another thing we should be aware of: This fix can introduce non-ASCII characters to currently pure ASCII txts. Especially English song creators usually didn't have to worry about encoding until now and may run into issues when uploading UTF-8 to USDB.

If I understand correctly, ASCII is not relevant here, is it? CP1252 contains the typographer’s apostrophe (`'`, U+2019, `RIGHT SINGLE QUOTATION MARK`, https://en.wikipedia.org/wiki/Right_single_quotation_mark) at position 92.

From the Wikipedia page: "The [Unicode](https://en.wikipedia.org/wiki/Unicode) character ’ (U+2019 right single quotation mark) is used both for a [typographic apostrophe](https://en.wikipedia.org/wiki/Apostrophe#Typographic_form) and a single right (closing) [quotation mark](https://en.wikipedia.org/wiki/Quotation_mark)."

Thus, if the upright apostrophe `'` IS used as a quotation mark and we replace it with `’`, then it is at least (more) correct for the closing quotation mark. Only the opening quotation mark `'` should be replace by `‘`, but instead is also replaced by the closing quotation mark/typographer’s apostrophe `’ `.

RumovZ:
Only the bytes 00 through 7F are ASCII. Note that 92 is a hexadecimal number, so while `’` is included in CP1252, it is encoded differently than its UTF-8 equivalent:

```py
>>> "’".encode("utf8")
b'\xe2\x80\x99'
>>> "’".encode("cp1252")
b'\x92'
```

However, I may have been wrong about "issues when uploading UTF-8 to USDB". As USDB understands UTF-8 and only enforces CP1252 when writing, `’` should work fine.
Karaoke clients are rather what could cause trouble here. A pure ASCII file can be read by any program, but `’` will be rendered incorrectly unless a client-compatible encoding has been used.

bohning:
I understand and agree. But it could be a feature to enforce the use of correct encodings for the target Karaoke software. (e.g. FAQ: Q: "Why are all my apostrophes wrong?" A: "Select the correct encoding in Settings when downloading the text file.").

bohning:
And with #42 , it is even less likely that a wrong target encoding is selected.

bohning:
Maybe a fix_apostrophes() function should be introduced which in turn calls both `self.notes.fix_apostrophes()` and `self.headers.fix_apostrophes()`?

src/usdb_syncer/notes_parser.py
```diff
@@ -252,6 +252,14 @@ def fix_pitch_values(self) -> None:
             for note in self.all_notes():
                 note.pitch = note.pitch - octave_shift * 12
 
+    def fix_apostrophes(self) -> None:
+        for note in self.all_notes():
+            note.text = (
+                note.text.replace("''", '"')  # two single quotation marks '' by "
+                .replace("`", "’")  # grave accent ` by upright apostrophe '
+                .replace("´", "’")  # acute accent ´ by upright apostrophe '
```
RumovZ, line 260: This is still `’`.
To avoid duplication, I would use a function here:
```py
def replace_false_apostrophes_and_quotation_marks(value: str) -> str:
    return value.replace("''", '"')  # two single quotation marks '' by "
            .replace("`", "'")  # grave accent ` by upright apostrophe '
            .replace("´", "'"),  # acute accent ´ by upright apostrophe '
```

bohning: Good catch! I overlooked that one. Thanks for the suggestion with the function, I added another commit.

src/usdb_syncer/notes_parser.py
```diff
@@ -384,6 +388,21 @@ def __str__(self) -> str:
     def artist_title_str(self) -> str:
         return f"{self.artist} - {self.title}"
 
+    def fix_apostrophes(self) -> None:
+        for key in (
+            "artist",
+            "title",
+            "language",
+            "genre",
+            "creator",
```
RumovZ, line 397: Maybe we should skip this one? We can also skip `comment` since we don't use that field.

bohning: You mean skip title, language, genre, creator and comment? But accents used as apostrophes are always wrong. They could be used in all those tags.

RumovZ: No, sorry, I just meant creator. If someone chooses to be called `It´s me`, we probably want to reuse that name literallly, don't we?

bohning: Ah, I see. Yes, maybe we want to take the creator's name literally, even though it hurts my eyes ;-)

RumovZ (approved):

PR #79, opened 2023-01-11
Fix songs that are entirely ALL CAPS.

bohning:
Lyrics in ALL CAPS is considered as shouting and therefore discouraged/bad style. Thus, if a song’s lyrics are entirely (!) ALL CAPS, they are all changed to lowercase and then the first word (more precisely the first capitalizable letter of the first word) of each line is capitalized again.

The only downside is that names/nouns that should be capitalized are then lowercase, but this should only concern a handful of words, so fixing this manually only takes a small effort.

Good test cases are e.g.:
http://usdb.animux.de/index.php?link=detail&id=9417
http://usdb.animux.de/index.php?link=detail&id=18539
http://usdb.animux.de/index.php?link=detail&id=21467
http://usdb.animux.de/index.php?link=detail&id=22046

### DIFF
--- a/src/usdb_syncer/notes_parser.py
+++ b/src/usdb_syncer/notes_parser.py
@@ -254,6 +254,11 @@ class Tracks:
     def lyrics(self) -> str:
         return "".join(note.text for note in self.all_notes())
 
+    def is_all_caps(self) -> bool:
+        return not any(
+            char.islower() for note in self.all_notes() for char in note.text
+        )
+
     def fix_line_breaks(self) -> None:
         for track in self.all_tracks():
             fix_line_breaks(track)
@@ -304,7 +309,7 @@ class Tracks:
             line.notes[-1].right_trim_text_and_add_space()
 
     def fix_all_caps(self) -> None:
-        if all(note.text.isupper() for note in self.all_notes()):
+        if self.is_all_caps():
             for note in self.all_notes():
                 note.text = note.text.lower()
             self.fix_first_words_capitalization()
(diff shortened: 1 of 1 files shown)


src/usdb_syncer/notes_parser.py
```diff
@@ -300,6 +303,28 @@ def fix_spaces(self) -> None:
...
+                note.text = note.text.lower()
+            self.fix_first_words_capitalization()
+
+    def fix_first_words_capitalization(self) -> None:
+        for line in self.all_lines():
+            # capitalize first capitalizable character
+            # e.g. "“¿que horas son?”" -> "“¿Que horas son?”"
+            text = ""
+            for char_index, char in enumerate(line.notes[0].text):
+                if char == char.capitalize():
```
RumovZ, line 318: Slightly simpler:
```suggestion
                if char != char.upper():
                    line.notes[0].text = line.notes[0].text.replace(char, char.upper(), 1)
                    break
```

src/usdb_syncer/notes_parser.py
```diff
@@ -300,6 +303,28 @@ def fix_spaces(self) -> None:
             # used to be incomplete in USDX, and it allows simple text concatenation
             line.notes[-1].right_trim_text_and_add_space()
 
+    def fix_all_caps(self) -> None:
+        if self.lyrics().isupper():
```
RumovZ, line 307: You don't need to construct a new string here:
```suggestion
        if all(note.text.isupper() for note in self.all_notes()):
```

bohning:
Thanks for the improvement suggestions!

RumovZ:
Sorry, that wasn't meant to be applied verbatim. There are other lines that need to be deleted now.

bohning:
Oh, I see. The convenience to commit the suggested improvements with a single click was too tempting =)

bohning:
Problem is: this code fails on the given song, because the very last syllable is (erroneously) on a single `!`. `isupper()` returns False on the exclamation mark, therefore the `if all()` expression returns False and thus the code is never executed.

RumovZ:
Oh well, I guess case is more complex than it may seem at a first glance. In that case, I would check individual letters:
```py
def is_all_caps(self) -> bool:
    return not any(char.islower() for note in self.all_notes() for char in note.text)
```

RumovZ (approved):

PR #81, opened 2023-01-12
First implementation of fix_relative_songs().

bohning:
I’ve adapted the implementation of UltraStar Manager (see: https://github.com/UltraStar-Deluxe/UltraStar-Manager/blob/d0f7e879d856f487ae0abed9d484c615f13ae1ee/src/plugins/lyric/QULyricTask.cpp#L797)

To test this, I have temporarily converted http://usdb.animux.de/index.php?link=detail&id=25177 to relative timings.


src/usdb_syncer/notes_parser.py
```diff
@@ -575,6 +576,30 @@ def minimum_song_length(self) -> str:
...
+                note.start = note.start + offset  # type: ignore
+
+            if line.line_break and line.line_break.out_time and line.line_break.in_time:
+                line.line_break.out_time = line.line_break.out_time + offset  # type: ignore
+                line.line_break.in_time = line.line_break.in_time + offset  # type: ignore
+
+                offset = line.line_break.in_time
+
+                if line.line_break.out_time == line.line_break.in_time:
+                    line.line_break.in_time = None
```
RumovZ, line 598: We shouldn't use `# type: ignore` too freely, because it could obscure real errors. In this case, there only seems to be a limitation with pylace and the comments are not required for our checks to pass, so I would rather not add them.
Anyway, I found a solution that is also understood by pylance. Please check that the logic is still correct.

src/usdb_syncer/notes_parser.py
```diff
@@ -85,8 +85,8 @@ class LineBreak:
     the next line start.
     """
 
-    start: int
-    end: int | None
+    out_time: int
+    in_time: int | None
```
RumovZ, line 89: I find this naming less intuitive. I would expect attributes to refer to the class they belong to, in this case out time to be the end of the line break. However, I admit the previous names weren't ideal, either.
Any other ideas? Maybe `last_line_out_time` and `next_line_in_time`?

RumovZ:
This module is getting quite large. Should we make it a package named `song_txt` and move tracks and headers into their own modules? By the way, let me know if you have a better name than `SongTxt`.

src/usdb_syncer/notes_parser.py
bohning: I fully agree, but tbh I did not understand why pylance complained operator `+` not being available for int and int | None, the latter referring to offset which is clearly initialized to 0 in line 583. So this was a temporary workaround to be able to submit the PR for further discussion.
I just checked, the code still works as expected. It is essential that relative songs are fixed before calling `maybe_split_duet_notes()`, otherwise this will fail on relative songs. So maybe `maybe_split_duet_notes() should double check that the song does not have relative timings.

src/usdb_syncer/notes_parser.py
bohning: The new naming came from UltraStar Manager and I find it more understandable than start/end, but possibly still not ideal. I am okay with your suggestion, even though then it should be `current_line_out_time` and `next_line_in_time`, right?

bohning:
> This module is getting quite large. Should we make it a package named `song_txt` and move tracks and headers into their own modules? By the way, let me know if you have a better name than `SongTxt`.

Sure, please feel free to refactor into a new package if that makes more sense.

src/usdb_syncer/notes_parser.py
RumovZ: My best guess is that it was some kind of circular issue with `offset` and `in_time` setting each other to `int | None` due to the lines `offset = line.line_break.in_time` and `line.line_break.in_time = line.line_break.in_time + offset`.

src/usdb_syncer/notes_parser.py
RumovZ: You mean because `LineBreak` is part of a `Line`? I'd have said that's more of an implementation detail, but either works for me.
Maybe strictly speaking, each line should have an `in_time` and `out_time` instead. But this might be more work than it's worth.

bohning: Yeah, true. From the point of view of the line_break, looking back it `last_line_out_time` (or `previous_line_out_time`) and looking forward it’s `next_line_in_time`. Either one is fine for me as well. I get your last point and agree that it’s more work than it’s worth.

RumovZ (approved):
🚀

bohning:
Super, vielen Dank!

Issue #73, opened 2023-01-06
redirecting bg URL in Meta-Tag saves invalid jpg with html content, does not log properly and throws exception


mjhalwa:
## Problem

if the `bg` Tag contains a URL that is redirecting to a different page, the generated background image instead contains the html content of the redirected page. In the current scenario the target URL is not an image itself but a webpage.

The syncer UI does not log anything about the image. The console output shows a Python traceback -> exception seems not to be caught.

## Expected

Either the background image is downloaded or the syncers log should indicate that the url is invalid (probably also how it is invalid - in this case leading to a webpage not an image). If the background image cannot be loaded a file should not be generated.

## Details:

**WARNING: Following the stated problematic URL in `bg` redirects leads to PORNOGRAPHIC webpage!**

Currently there are 3 songs for "Disney's Eiskönigin - Lass jetzt los"

- http://usdb.animux.de/index.php?link=detail&id=17227
- http://usdb.animux.de/index.php?link=detail&id=18408
- http://usdb.animux.de/index.php?link=detail&id=18719

which have the same meta-tag:

``` conf
#VIDEO:v=Jc10TXOh-yM,co=static.universal-music.de/asset_new/480408/195/view/Die-Eiskoenigin---Voellig-unverfroren--Deutscher-Original-Soundtrack.jpg,bg=motaen.com/upload/wallpapers/source/2014/01/26/20/04/39018/tNXdnd2Xtx.jpg 
```

The complete log from within the syncer (syncer start and only download of 17227) is:

``` log
2023-01-06 22:12:16 [INFO] Application successfully loaded.
2023-01-06 22:12:38 [INFO] #17227: Found 'Disney’s Die Eiskönigin - Lass jetzt los' on  USDB
2023-01-06 22:13:04 [INFO] #17227: Success! Downloaded audio.
2023-01-06 22:13:19 [INFO] #17227: Success! Downloaded video.
2023-01-06 22:13:20 [INFO] #17227: Success! Downloaded cover.
```

end of log on the console:

``` log
#...
2023-01-06 22:13:19 [INFO] #17227: Success! Downloaded video.
2023-01-06 22:13:20 [INFO] #17227: Success! Downloaded cover.
Traceback (most recent call last):
  File "{PROJECTPATH}\usdb_syncer\src\usdb_syncer\song_loader.py", line 157, in run
    _maybe_download_background(ctx)
[...]

PR #82, opened 2023-01-12
check filetype after download and remove non-matching files with error

mjhalwa:
In order to address #73 a first layer of protection may be to check file types after download to identify non-image/audio/video files and delete them. Additionally it offers the benefit of faster identification of broken links that need to be fixed.

__!! requires additional python package !!__
  - `filetype`: [PyPI](https://pypi.org/project/filetype/)
    > Small and dependency free [Python](http://python.org/) package to infer file type and MIME type checking the [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)#Magic_numbers_in_files) signature of a file or buffer. 

Do I miss a step to correctly add a new package or are taken step sufficient?

After download of images (cover / background) and videos (audio/video) the package `filetype` is used to check the files being image or audio/video. (Our m4a-files are identified as video/mp4, but the check is in `download_video()` which downloads both so should be fair enough. Any missmatching files are deleted and an error is issued to the logger.


mjhalwa:
I tried to find some broken links by trying to cherry pick songs with special cover-/background-urls, but gave up at about USDB ID 602, without any luck.

src/usdb_syncer/resource_dl.py
```diff
@@ -96,7 +97,14 @@ def download_video(
             logger.debug(f"error downloading video url: {url}")
             return None
 
-    return ext or os.path.splitext(filename)[1][1:]
+    actual_file_ext = ext or os.path.splitext(filename)[1][1:]
+    path = f"{path_stem}.{actual_file_ext}"
+    if not (filetype.is_video(path) or filetype.is_audio(path)):
```
RumovZ, line 102: Is this necessary? I can't imagine yt_dlp works with anything but valid video URLs.

RumovZ (changes requested):
Cool, this feature makes a lot of sense!
You don't need a real txt to test it. You could add something like `ctx.txt.meta_tags.cover = ImageMetaTags("example.com")` in `song_loader.py:227`.

src/usdb_syncer/resource_dl.py
```diff
@@ -142,6 +150,12 @@ def download_and_process_image(
     path = os.path.join(pathname, fname)
     with open(path, "wb") as file:
         file.write(img_bytes)
+
+    if not filetype.is_image(path):
```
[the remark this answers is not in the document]
mjhalwa: ah thanks, did not think of that. I will check out how to test the bytes instead of a file

src/usdb_syncer/resource_dl.py
mjhalwa: to be honest I just put the image check in the image download function and did the same for audio/video. But it may be useful if yt_dlp left an empty file after an exception or video/audio post processing had an error and ceated a wrong formatted / empty file.

RumovZ: But your code isn't even executed in case of an exception. And even if you moved it into the except block and yt_dlp did leave any files behind, you wouldn't know which path to check.
In summary, cleaning up is really yt_dlp's responsibility here and it seems to do that very reliably.

mjhalwa: oh true, did not register the return in the except block. I agree and will remove the check here

mjhalwa:
applied both changes

RumovZ (approved):

RumovZ:
Nice, thanks!

PR #88, opened 2023-01-28
Normalize audio using ffmpeg_normalize.

bohning:
This seems to accomplish what is needed, but in a more concise way than two external calls to ffmpeg.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -28,6 +28,7 @@ install_requires =
     attrs
     pyshorteners
     filetype
+    ffmpeg-normalize
 
 [options.packages.find]
 where = src
--- a/src/usdb_syncer/download_options.py
+++ b/src/usdb_syncer/download_options.py
@@ -19,6 +19,7 @@ class AudioOptions:
 
     format: settings.AudioFormat
     bitrate: settings.AudioBitrate
+    normalize: bool
 
     def ytdl_format(self) -> str:
         return self.format.ytdl_format()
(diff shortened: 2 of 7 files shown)


bohning:
I am just a little bit sceptic that the output of ffmpeg-normalize states `normalization_type: dynamic`, which is contradictory to the manual.
image

mjhalwa:
As discussed in voice chat: This message seems to be an output from the first pass. I tested the proposed normalization with 3 of my songs (#65 ) + "Monsterparty". The comparison with my dynamic and linear results unveiled that the proposed method with ffmpeg_normalize does in fact produce linearly normalized results - even though the output states "dynamic". One of my samples (Circle of Life) even has an LRA of 16.2 and still was linearly normalized.

bohning:
Open issues:
- importing ffmpeg-normalize messes up our TextEditLogger due to introduced log coloring (should be solved upstream in ffmpeg-normalize, @g3n35i5 offered to have a look at it)
- if normalization is checked and AudioFormat = M4A, we should probably drop the `ext=m4a` from the yt-dlp command so that it also downloads the .webm (which is then normalized and converted to an m4a in the process, as is the case when AudioFormat = MP3).

g3n35i5:
> * importing ffmpeg-normalize messes up our TextEditLogger due to introduced log coloring (should be solved upstream in ffmpeg-normalize, @g3n35i5 offered to have a look at it)

https://github.com/slhck/ffmpeg-normalize/pull/216

g3n35i5:
Ffmpeg normalize PR ist gemerged

bohning:
> Ffmpeg normalize PR ist gemerged

Works like a charm now and does not mess up our own logging anymore.

RumovZ (approved):

Issue #69, opened 2022-12-31
Miscellaneous UI improvements


BWagener:
I wanted to start a discussion on UI improvements.
There are two things I have noticed so far that could be improved in my opinion:

- main window size exceeds smaller screens
  - this is how the syncer looks on a 1280x800 screen (full screen): 
![image](https://user-images.githubusercontent.com/19321042/210140879-ffd63824-2080-4fca-b9af-71a44801f80b.png)
  - this is how the syncer looks on a 1280x800 screen (docked to the side - half screen as a workaround): 
![image](https://user-images.githubusercontent.com/19321042/210140975-a5003901-51e1-479e-b6e3-0d698d7af0a3.png)
  - this is how it looks with an [attempted improvement](https://github.com/bohning/usdb_syncer/compare/main...BWagener:usdb_syncer:fix_filter_width): ![image](https://user-images.githubusercontent.com/19321042/210140784-846308e9-efab-4271-b5c9-7b90bd1e063d.png)
    - I have assigned a fixed width to the filter combo boxes to achieve this, this is not a nice solution but mitigates the combo boxes always assuming the width of their longest item, which are quite long for the artist and title boxes. This wouldn't look nice on higher resolutions.
- the song table exceeds the main window (see third screenshot)
  - the table columns assume the width of their longest element if their header is set to `QHeaderView.ResizeMode.Interactive`, which due to a few very long artist and title names makes the table exceed the main window
  - a solution to this could be defaulting the interactive columns to assume a fraction of the table parent's width (minus the fixed columns width). At least up to a minimum width, below which the table should still be allowed to exceed it's parent horizontally
  - another nice addition on this front would be to remember user configured column width and default to those if they exist, I found [this approach](https://www.qtcentre.org/threads/26675-QTableView-saving-restoring-columns-widths) to this

Let me know what you think about these points and any you have to add! I understand if certain improvements or additions are too early at this point in development but it still might be worth it to collect them and maybe split more important ones off into separate issues.

RumovZ:
Good points! On WSL, the app even crashes, because we try to open it in fullscreen, which doesn't fit.

I think the table is not that much of a problem, because it is a scroll area. But I agree that we waste a lot of space by adjusting to the longest entry. We probably don't need to worry too much about defaults if we remember user resizings. I would only hard code the width of the checkmark columns, which we statically know the maximum width of.
There are more columns we can potentially add in the future, so the ability to toggle columns may also be valuable.

[...]

PR #94, opened 2023-02-04
Fix filter width, use enums to identify table columns

BWagener:
This PR currently contains two changes, though both are just suggestions so please let me know if you think they make sense or should be changed or removed.

The first change is setting a minimum width on the filter combo boxes. This does not affect how the combo boxes are displayed on larger screens but on smaller screens this allows them to shrink and this in turn allows the entire window to fit smaller screens. This problem is outlined in #69.

The second change introduces the use of IntEnums for the table columns instead of referencing them using their index. Currently the ID column width is dynamically derived from the font width.

### DIFF
--- a/src/usdb_syncer/gui/forms/MainWindow.ui
+++ b/src/usdb_syncer/gui/forms/MainWindow.ui
@@ -98,6 +98,12 @@
     </item>
     <item>
      <widget class="QGroupBox" name="groupBox_4">
+      <property name="sizePolicy">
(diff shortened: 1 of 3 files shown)


BWagener:
If you consider some of these changes worth integrating let me know if I should squash the commits please.

RumovZ (approved):
That's already much better! Thanks!
And sorry for not coming back on this sooner. I still had this PR in my inbox as a reminder... I'm glad you stuck with it and made it mergeable!

PR #104, opened 2023-02-28
Remove trailing dots (and resulting spaces) in filenames.

bohning:
Technically only needed for Windows, but it is always done in order to support OS-independent song collections.

### DIFF
--- a/src/usdb_syncer/utils.py
+++ b/src/usdb_syncer/utils.py
@@ -67,6 +67,8 @@ def sanitize_filename(fname: str) -> str:
     for old, new in FILENAME_REPLACEMENTS:
         for char in old:
             fname = fname.replace(char, new)
+    if fname.endswith("."):
+        fname = fname.rstrip(" .")  # Windows does not like trailing periods
     return fname
 
 


bohning:
> Aren't we always using valid file extensions so there shouldn't be any trailing dots? And why are dots not proceded by a space left alone?

The problem arose from "AC/DC - T.N.T." folder. If the syncer tries to create this folder on Windows, the trailing period is silently removed by the OS, resulting in a folder named "AC/DC - T.N.T". Subsequently, the Syncer could not write to the folder.
Removing the trailing dot helps if there is one trailing dot as in the above example, but then there is "Bela B feat. Charlotte Roche - 1. ... 2. ... 3. ..." (in the meantime replaced by ellipses in USDB which seems to work for both Windows and Mac and thus probably only Linux) where all trailing dots need to be removed, leaving a trailing space preceded by a trailing dot, so I thought it could be solved by deleted all trailing spaces and periods.

Maybe there is a better solution?!

RumovZ:
Oh, I missed that we also use this for folders. The other point was wrong, too. Sorry, I should have thought a moment longer about this. 😅

RumovZ (approved):

PR #106, opened 2023-02-28
Now that USDB supports UTF8, the fixing of encodings is obsolete.

bohning:


RumovZ (approved):
🎉

bohning:
Sad to see this code go, it was a fantastic addition to the Syncer. At the same time I am happy about the upgrades in USDB.

PR #107, opened 2023-03-13
Detect if the user is not logged in and abort the download.

bohning:


src/usdb_syncer/song_loader.py
```diff
@@ -82,8 +82,10 @@ class Context:
     @classmethod
     def new(
         cls, details: SongDetails, options: Options, info: DownloadInfo, logger: Log
-    ) -> Context:
-        txt_str = usdb_scraper.get_notes(details.song_id, logger)
+    ) -> Union[Context, None]:
```
RumovZ, line 85: In Python 3.10 you generally don't need `Union`. You can write `Context | None` instead.

src/usdb_syncer/usdb_scraper.py
```diff
@@ -410,9 +410,11 @@ def get_notes(song_id: SongId, logger: Log) -> str:
...
+    text, success = _parse_song_txt_from_txt_page(soup)
+    return text, success
 
 
-def _parse_song_txt_from_txt_page(soup: BeautifulSoup) -> str:
-    return soup.find("textarea").string  # type: ignore
+def _parse_song_txt_from_txt_page(soup: BeautifulSoup) -> tuple[str, bool]:
+    if textarea := soup.find("textarea"):
+        return textarea.string, True  # type: ignore
+    return "", False
```
RumovZ, line 420: Can't we return `None` or an empty string in the error case here? I find that more readable and ergonomic than types you have to unpack first.

RumovZ (approved):
👍

src/usdb_syncer/song_loader.py
bohning: Thanks for the suggestion, I didn’t know that.

src/usdb_syncer/usdb_scraper.py
bohning: Always appreciate your improvement suggestions!

bohning:
I was wondering if there is a way to stop all queued downloads once the status of a missing login is determined.

RumovZ:
It's possible, but I think it would be messy. Maybe instead you could check before any download is started at all, i.e. when the button is clicked?

bohning:
Not sure if I did it the best way, but it seems to be working well.

RumovZ:
Looks good. 🙂

PR #111, opened 2023-03-24
Upload to usdb (for mods and admins)

bohning:
The idea is to (batch) upload a song right away after the download with all the Syncer fixes in place.

I am unsure what is the best way to reconstruct the meta-tags. Except for the medley tags, they should be unchanged, while the medley tags might need to be updated if the song’s BPM was changed by the Syncer.

For now, I used the __str__() of the MetaTags class, but it doesn’t feel right and there sure is a better way. Looking forward to your feedback and help.

### DIFF
--- a/src/usdb_syncer/usdb_scraper.py
+++ b/src/usdb_syncer/usdb_scraper.py
@@ -228,9 +228,7 @@ def get_usdb_mod_status() -> bool:
     html = get_usdb_page("index.php", params={"link": "home"})
     soup = BeautifulSoup(html, "lxml")
     usdb_strings = _get_usdb_strings(soup)
-    if usdb_strings.VERIFY_NEW_SONGS in soup.get_text():
-        return True
-    return False
+    return usdb_strings.VERIFY_NEW_SONGS in soup.get_text()
 
 
 def _parse_song_page(soup: BeautifulSoup, song_id: SongId) -> SongDetails:


src/usdb_syncer/meta_tags/deserializer.py
```diff
@@ -156,6 +156,45 @@ def __init__(self, video_tag: str, logger: Log) -> None:
                 case _:
                     logger.warning(f"unknown key for meta tag: '{pair}'")
 
+    def __str__(self) -> str:
```
RumovZ, line 159: There is already a serializer for meta tags which we use in the meta tags dialog. However, I have a local branch that refactors that code, so we should probably merge that in first. Only I cannot push it until next weekend, unfortunately.

src/usdb_syncer/song_loader.py
```diff
@@ -290,6 +297,44 @@ def _write_sync_meta(ctx: Context) -> None:
     ctx.sync_meta.to_file(ctx.locations.dir_path)
 
 
+def _maybe_upload_txt(ctx: Context) -> None:
+    if not get_usdb_mod_status():
```
RumovZ, line 301: Making an additional HTTP request for every user and every download seems excessive when it can't work for 99.9% of users anyway. I would use an environment variable instead. And if it's set, can't we just make the post request directly without checking the login status first?

src/usdb_syncer/usdb_scraper.py
```diff
@@ -224,6 +224,15 @@ def get_usdb_login_status() -> bool:
     return True
 
 
+def get_usdb_mod_status() -> bool:
+    html = get_usdb_page("index.php", params={"link": "home"})
+    soup = BeautifulSoup(html, "lxml")
+    usdb_strings = _get_usdb_strings(soup)
+    if usdb_strings.VERIFY_NEW_SONGS in soup.get_text():
+        return True
+    return False
```
RumovZ, line 233:
```suggestion
    return usdb_strings.VERIFY_NEW_SONGS in soup.get_text()
```

src/usdb_syncer/meta_tags/deserializer.py
bohning: Great, then let’s wait until that branch is merged and then we continue from there.

bohning:
Superseded by #479 .

PR #112, opened 2023-03-27
Normalize Unicode of USDB pages

### LINKED ISSUE #110: next_unique_directory() renames existing folder if it contains special characters (e.g. Umlauts)
Most likely, this is a unicode normalization issue. For example, "Maná" can be written as `Mana\\u0301` or `Man\\u00e1`. The behavior of Windows and Mac seems to differ here, where Mac seems to normalize the folder name while Windows doesn’t.

RumovZ:
Closes #110.
Does this work for you, @bohning? You probably have to delete the problematic folder and then restart the syncer.


bohning:
> Closes #110. Does this work for you, @bohning? You probably have to delete the problematic folder and then restart the syncer.

Yes, works perfectly. Thanks for the quick fix!

bohning (approved):

PR #114, opened 2023-03-28
Add function to fix overlapping notes and reorder fixes.

bohning:


RumovZ (commented):
Wouldn't it make sense to merge this and `fix_touching_notes()` into a single function?
Also, `Note.shorten()` could be changed to shorten to a minimum of 1 beat, instead of returning early when `self.duration >= beats`.

bohning:
> Wouldn't it make sense to merge this and `fix_touching_notes()` into a single function?
> Also, `Note.shorten()` could be changed to shorten to a minimum of 1 beat, instead of returning early when `self.duration >= beats`. (Or something more complex, if you want to avoid shortening notes that would still overlap anyway.)

Yes, maybe I haven't fully thought this through. My reasoning was that fix-overlapping at least shortens the earlier note so that they are touching, and fix-touching further shortens by 1 beat *if possible*.

But.yes, probably this can be simplified into one function that fixes both overlapping and touching notes.

RumovZ (approved):

PR #118, opened 2023-04-07
Sync resources individually dropping txt hash

RumovZ:


bohning (approved):

bohning:
In resource-dl.py, there is a typo: url_from_resouce is missing an 'r'

Issue #31, opened 2022-11-22
Offer options to only download certain song types


bohning:
Some songs in usdb already contain meta-tags, some can be completed by parsing the video links in the comments, yet for others only the text file (and maybe the small cover) can be downloaded.

The user should be able to select between the three options:
1. Only songs that contain meta-tags
2. Only songs with either meta-tags and/or video/audio resources in the comments
3. All songs

Issue #16, opened 2022-11-15
Make folder structure configurable/flexible


bohning:
Currently, the folder structure is hardcoded `Artist - Title\{usdb_id}`. The user may wants to have a different folder structure, e.g. have a top level folder for every beginning letter of the Artist, or for every language or combinations of that. If we make this configurable, the sync process has to respect that directory structure.

Baklap4:
Really would love this feature!
Personally i store everything like this:
```
karaoke
|---->USDB-live
|------> English
|---------> ABBA
|-------------> ABBA - Mamma mia
|----------------> ABBA - Mamma mia.mp3, ABBA - Mamma mia.mp4, ABBA - Mamma mia.txt
|------> Dutch
|---------> Youp van 't Hek
|-------------> Youp van 't Hek - Flappie
|----------------> Youp van 't Hek - Flappie.mp3, Youp van 't Hek - Flappie.mp4, Youp van 't Hek - Flappie.txt
```

But making it configurable would make even more sense :)

bohning:
Unfortunately not. It's on the roadmap, but tbh with a low priority.

bohning:
But it is already possible to reorganize your collection as you like, as long as you restart the syncer. On restart, the syncer will keep track of where to find which song.

RumovZ:
Closed with #270.

Baklap4:
Thanks going to check it out somewhere next week :D!

PR #123, opened 2023-04-20
Fix song check and golden notes localization and put SC songs into a separate subdirectory

bohning:


src/usdb_syncer/usdb_scraper.py
```diff
@@ -335,8 +335,8 @@ def _parse_details_table(
         cover_url=cover_url,
         bpm=details_table.find(string="BPM").next.text,  # type: ignore
         gap=details_table.find(string="GAP").next.text,  # type: ignore
-        golden_notes=details_table.find(string=usdb_strings.GOLDEN_NOTES).next.text,  # type: ignore
-        song_check=details_table.find(string=usdb_strings.SONGCHECK).next.text,  # type: ignore
+        golden_notes=details_table.find(string=usdb_strings.GOLDEN_NOTES).next.text.rstrip() == usdb_strings.YES,  # type: ignore
+        song_check=details_table.find(string=usdb_strings.SONGCHECK).next.text.rstrip() == usdb_strings.YES,  # type: ignore
```
RumovZ, line 339: I guess black doesn't complain, because it can't refactor these lines due to the comment, but they are still too long.
Maybe we could add a type safe helper for the `details_table.find(string=...).next.text` part, which comes up a lot.

src/usdb_syncer/song_loader.py
```diff
@@ -48,10 +48,17 @@ class Locations:
...
+        song_id: SongId,
+        song_check: bool,
+        song_dir: Path,
+        meta_path: Path | None,
+        headers: Headers,
     ) -> Locations:
         filename_stem = sanitize_filename(headers.artist_title_str())
         if not meta_path:
+            if song_check:
+                song_dir = song_dir.joinpath("[SC]-Songs")
```
RumovZ, line 61: Maybe we should delay this until #16? I personally prefer less subfolders.

src/usdb_syncer/usdb_scraper.py
bohning: What is a type safe helper?

src/usdb_syncer/song_loader.py
bohning: Sure, maybe also related to #31.

RumovZ (approved):

PR #126, opened 2023-04-28
Add id3/meta tags to audio file. Closes #113.

### LINKED ISSUE #113: Generate suitable MP3 Tags
It would be nice if the resulting files after a sync contained suitable MP3Tags to work with Tools like UltraStar Manager.

bohning:
This PR adds the available artist/title/language/genre/year/lyrics information and, if selected in the preferences, the available artworks to the audio file.


src/usdb_syncer/song_loader.py
```diff
@@ -339,3 +343,88 @@ def _maybe_write_txt(ctx: Context) -> None:
...
+
+    match audiofile.suffix:
+        case ".m4a":
+            _write_m4a_tags(meta, ctx, options.embed_artwork)
+        case ".mp3":
+            _write_mp3_tags(meta, ctx, options.embed_artwork)
+
+
+def _write_m4a_tags(audio_meta: FileMeta, ctx: Context, embed_artwork: bool) -> None:
+    path = ctx.locations.file_path(audio_meta.fname)
```
RumovZ, line 362: Shouldn't this be removed as in `_write_mp3_tags()`?

src/usdb_syncer/song_loader.py
```diff
@@ -339,3 +343,88 @@ def _maybe_write_txt(ctx: Context) -> None:
...
+    m4a = mutagen.mp4.MP4(path)
```
RumovZ, line 363: Shouldn't `path` be removed as in `_write_mp3_tags()`? Below you check `if m4a.tags`. Does this mean no tags will be written if the file does not already have tags?

bohning: It seems that `mutagen.mp4.MP4()` needs an actual file as parameter (there is no equivalent to ID3() for m4a files) and `m4a = mutagen.mp4.MP4()` does not work, while `tags = id3.ID3()` creates an empty ID3 object, then fills it and then writes it to the file. Not sure if `if m4a.tags` is needed, but if I leave it out, I get `Unsupported target for indexed assignment ("None")  [index]mypy(error)` errors.

RumovZ: Ok, so the API requires it for whatever reason.
As for the second point, why not create and assign a new `MP4Tags` instance?

bohning: Good point. I can try if that works as well.

bohning: Seems to work. Just pushed the changes.

RumovZ (approved):

PR #141, opened 2023-07-16
Fix staging local song matches

RumovZ:
Was staging the wrong songs, because indices were taken from the proxy model.
This should also be faster and scale better.


bohning (approved):
Awesome!

PR #152, opened 2023-08-27
Fix consecutive_notes() for more than one track

RumovZ:


bohning (approved):
Thanks for the quick fix!

PR #159, opened 2023-09-04
use pipenv, Pipfile and Pipfile.lock to have more control over dependencies

mjhalwa:
new setup procedure:

```
pipenv install --dev
```

new execution procedure

```
pipenv run usdb_syncer
```

instead of `./venv` creates `./.venv`

### DIFF
--- a/.env
+++ b/.env
@@ -1,0 +1,2 @@
+# pipenv configs
+PIPENV_VENV_IN_PROJECT=1
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+Pipfile       text eol=lf
+Pipfile.lock  text eol=lf
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -37,22 +37,24 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: "3.11"
+      - name: Install Pipenv
+        run: |
+          python -m pip install pipenv
       - name: Install dependencies
         run: |
-          python -m pip install --upgrade pip wheel
-          python -m pip install '.[dev]'
+          pipenv install --dev --deploy
       - name: Write version and commit hash
-        run: write_release_info -p src/usdb_syncer/constants.py -v ${{ github.ref_name }} -c ${{ github.workflow_sha }}
+        run: pipenv run write_release_info -p src/usdb_syncer/constants.py -v ${{ github.ref_name }} -c ${{ github.workflow_sha }}
       - name: Generate GUI elements
-        run: generate_pyside_files
+        run: pipenv run generate_pyside_files
       - name: Scrape USDB song list to ship with the bundle
-        run: generate_song_list_json -t 'data/song_list.json' -u '${{ secrets.USDB_USER }}' -p '${{ secrets.USDB_PASSWORD }}'
+        run: pipenv run generate_song_list_json -t 'data/song_list.json' -u '${{ secrets.USDB_USER }}' -p '${{ secrets.USDB_PASSWORD }}'
       - name: add dependencies for xcb support
         if: matrix.TARGET == 'Linux'
         run: |
           sudo apt install libxcb-cursor0
       - name: Build with pyinstaller for ${{matrix.os}}
-        run: pyinstaller -n 'USDB_Syncer-${{github.ref_name}}-${{matrix.TARGET}}' ${{matrix.PYINSTALLER_ARGS}} src/usdb_syncer/main.py
+        run: pipenv run pyinstaller -n 'USDB_Syncer-${{github.ref_name}}-${{matrix.TARGET}}' ${{matrix.PYINSTALLER_ARGS}} src/usdb_syncer/main.py
       - name: Create DMG
         if: matrix.TARGET == 'macOS'
         run: |
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -42,19 +42,19 @@ jobs:
           python -m pip install pipenv
       - name: Install dependencies
         run: |
-          pipenv install --dev --deploy
+          pipenv sync
       - name: Write version and commit hash
-        run: pipenv run write_release_info -p src/usdb_syncer/constants.py -v ${{ github.ref_name }} -c ${{ github.workflow_sha }}
+        run: python -m src.tools.write_release_info -p src/usdb_syncer/constants.py -v ${{ github.ref_name }} -c ${{ github.workflow_sha }}
       - name: Generate GUI elements
-        run: pipenv run generate_pyside_files
+        run: python -m src.tools.generate_pyside_files
       - name: Scrape USDB song list to ship with the bundle
-        run: pipenv run generate_song_list_json -t 'data/song_list.json' -u '${{ secrets.USDB_USER }}' -p '${{ secrets.USDB_PASSWORD }}'
+        run: python -m src.tools.generate_song_list_json -t 'data/song_list.json' -u '${{ secrets.USDB_USER }}' -p '${{ secrets.USDB_PASSWORD }}'
       - name: add dependencies for xcb support
         if: matrix.TARGET == 'Linux'
         run: |
           sudo apt install libxcb-cursor0
       - name: Build with pyinstaller for ${{matrix.os}}
-        run: pipenv run pyinstaller -n 'USDB_Syncer-${{github.ref_name}}-${{matrix.TARGET}}' ${{matrix.PYINSTALLER_ARGS}} src/usdb_syncer/main.py
+        run: pyinstaller -n 'USDB_Syncer-${{github.ref_name}}-${{matrix.TARGET}}' ${{matrix.PYINSTALLER_ARGS}} src/usdb_syncer/main.py
       - name: Create DMG
         if: matrix.TARGET == 'macOS'
         run: |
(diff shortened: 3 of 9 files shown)


mjhalwa:
you can already try out locally, if this works for you too. I still have to adapt the github action files to work with pipenv instead of setup.cfg.

mjhalwa:
ok so tox, linting and tests seem to work now... just need to check if release building works too, then this can be reviewed.

.env
```diff
@@ -0,0 +1,2 @@
+# pipenv configs
+PIPENV_VENV_IN_PROJECT=1
```
RumovZ, line 2: I don't really have a preference myself, but I'm curious, is there any reason to overwrite the default?

RumovZ, line 2: I don't really have a preference myself, but I'm curious, is there any reason to overwrite the default?

.gitattributes
```diff
@@ -0,0 +1,2 @@
+Pipfile       text eol=lf
+Pipfile.lock  text eol=lf
```
RumovZ, line 2: Probably not required, as the whole codebase is LF anyway.

.github/workflows/release.yaml
```diff
@@ -37,22 +37,24 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: "3.11"
+      - name: Install Pipenv
+        run: |
+          python -m pip install pipenv
```
RumovZ, line 42: We should probably use the `--system` flag here, as there's no need to use a virtual env on the VM.

README.md
```diff
@@ -80,22 +80,16 @@ pkg-config --cflags --libs dbus-1
...
-source venv/bin/activate
-pip install --upgrade pip tox setuptools
-pip install -e '.[dev]'
+pipenv install --dev
 ```
 
 ### Run usdb_syncer
 
-The package has a defined entry point for the GUI. Simply type in `usdb_syncer` in your terminal. Make sure that your venv is activated.
+The package has a defined entry point for the GUI. Simply type in `pipenv run usdb_syncer` in your terminal.
```
RumovZ, line 88: Wouldn't it be easier to activate the venv with `pipenv shell`? Then you don't have to change all the commands.
`pipenv run usdb_syncer` is also quite cumbersome to type, as there is no atuo completion. In the venv I can just type `u` + <kbd>Tab</kbd> or so.

tox.ini
```diff
@@ -8,7 +8,11 @@ python = 3.11: py311-{lint, test}
 [testenv:py311-lint]
 description = Run static checkers.
 extras = lint
+allowlist_externals=pipenv
 commands =
+    # install dependencies
+    pipenv install --dev --deploy
```
RumovZ, line 14: Not sure if it matters, but pipenv gives the following notice:
> Courtesy Notice: Pipenv found itself running within a virtual environment, so it will automatically use that environment, instead of creating its own for any project. You can set PIPENV_IGNORE_VIRTUALENVS=1 to force pipenv to ignore that environment and create its own instead. You can set PIPENV_VERBOSITY=-1 to suppress this warning.

If there's no more elegant way to do this, maybe we should use `PIPENV_VERBOSITY=-1` to explicitly acknowledge it.

RumovZ (commented):
Thanks for looking into this! After removing the old envs, it works great for me. Just a few minor questions below.

.env
mjhalwa: we always hat the venv inside the project directory, I just wanted to keep it as is. For me its easier that way to enter/delete the venv instead of having to looki it up with `pipenv --venv`.
Note: if you run `pipenv run python` you get a python console in the venv, but without arrow-up history functionality (which is broken for pipenv)

.gitattributes
mjhalwa: For me it was required, because git changed it on upload. It is crucial, that the Pipfile does not get changed, because the file hash of the Pipfile is stored in Pipfile.lock. When deploying pipenv with `--deploy` the hash is checked and the deployment fails on missmatch

.github/workflows/release.yaml
mjhalwa: Indeed I want to try that. I guess the build inside the venv is the reason that I encounter gbome_keyring errors on ubuntu currently. Also I would like to get rid of the `--dev` deployment. It should actually be productive. I thought of removing pyinstaller from the dev dependencies and install it in the action manual for installs.

The reason for doing it this way currently is that releases broke with the `--system` flag in my first tries.

README.md
mjhalwa: actially both is possible. Its my personal preference to invole the `pipenv run`-syntax, because I do not need to struggle with beeing inside the venv or not.
If you prefer, I can revert the text changes and add `pipenv shell` to end of the code block

tox.ini
mjhalwa: The problem here was that `tox` creates an own environment. I tried to adapt the config in order to make it work, but wirhout knowing mcuh about `tox`. Just one thing that comes to my mind: if we added `PIPENV_VERBOSITY=-1` to the `.env` file than the config would affect both the actual environment creation and the tox environment creation. Passing the environment variable on console in the `tox.ini` would be a different syntax on each OS (e.g. `$Env:PIPENV_VERBOSITY=-1` in powershell) wouldn't it?

.gitattributes
RumovZ: Sounds like you have set `autocrlf = true` in your .gitconfig. In my experience, that's only causing trouble, and modern text editors work just fine with any line endings you prefer. Just the thought of my VCS implicitly changing the code I commit! 😬

.github/workflows/release.yaml
RumovZ: Your comment above made me thinking, shouldn't we also use `sync` instead of `install` in CICD, and maybe even when setting up the local venv?

.env
RumovZ: Ok, I see.
For removing the venv `pipenv --rm` is handy.

README.md
RumovZ: Thanks, I would prefer that. Then we can keep using our entry points just like before.

tox.ini
RumovZ: We can add the env var to the tox environment only:
```diff
diff --git a/tox.ini b/tox.ini
index 02d374f..eb20c98 100644
--- a/tox.ini
+++ b/tox.ini
@@ -9,6 +9,8 @@ python = 3.11: py311-{lint, test}
 description = Run static checkers.
 extras = lint
 allowlist_externals=pipenv
+; suppress warning about pipenv running in a venv
+set_env = PIPENV_VERBOSITY=-1
 commands =
     # install dependencies
     pipenv install --dev --deploy
@@ -33,6 +35,8 @@ commands =
 description = Run doc tests and unit tests.
 extras = test
 allowlist_externals=pipenv
+; suppress warning about pipenv running in a venv
+set_env = PIPENV_VERBOSITY=-1
 commands =
     # install dependencies
     pipenv install --dev --deploy
```

At least, pipenv is smart enough to detect the venv and not create another one. I think that's good enough for now, but let's hope they improve support for each other in the future.

.env
mjhalwa: ah ya true, I forget about that. So should we keep it inside the project directory or not?

.gitattributes
mjhalwa: I need to check that. :sweat_smile:

.env
RumovZ: Sure, like I said, was just curious. 🙂

.github/workflows/release.yaml
mjhalwa: funny thing... `pipenv sync` also creates a virtual environment, that's why it still needs `pipenv sync --system` 🤷

RumovZ: Yes, in this regard, it works just like `install`. However, it doesn't update the pip or lock file. That's why I think it should be used in CI/CD, and definitely when releasing.
Sorry if I confused you, this wasn't meant as an answer to the venv issue.

See here: https://pipenv.pypa.io/en/latest/commands/

.github/workflows/release.yaml
```diff
@@ -37,16 +37,41 @@ jobs:
...
+        if: matrix.TARGET == 'Windows'
+      - name: Prepend src to PYTHONPATH (Linux/macOS)
+        run: |
+          echo "PERSISTENT_PYTHONPATH=$(pwd)/src:$PYTHONPATH" >> $GITHUB_ENV
+        if: matrix.TARGET == 'Linux' || matrix.TARGET == 'macOS'
+      - name: Write version and commit hash (Windows DEBUG)
+        run: |
+          echo $env:PYTHONPATH
+          echo $PYTHONPATH
+          python -m src.tools.write_release_info -p src/usdb_syncer/constants.py -v ${{ github.ref_name }} -c ${{ github.workflow_sha }}
```
RumovZ, line 58: Did you try without the `-m` flag?
```suggestion
        run: python ./src/tools/write_release_info.py -p src/usdb_syncer/constants.py -v ${{ github.ref_name }} -c ${{ github.workflow_sha }}
```
It should be possible without modifyng the PYTHONPATH.

mjhalwa: I did and had the same problem of Module 'tools' cannot be found

mjhalwa:
Seems to work now. I checked Releases on
- windows 11: **works**
- Ubuntu 23.04: gnome_keyring error
- Debian 12: after  `gpg --full-gen-key could select a key to create the requested kdewallet and **works**

But I have no macos to test this build. I guess if the macos build works we can finally merge this. The `gnome_keyring` seems to be the known issue and unrelated to using pipenv to create a release for linux.

RumovZ (approved):

PR #160, opened 2023-09-18
Add song_id to song related log messages.

bohning:


src/usdb_syncer/usdb_scraper.py
```diff
@@ -287,14 +287,16 @@ def get_usdb_details(song_id: SongId) -> SongDetails:
     html = get_usdb_page(
         "index.php", params={"id": str(int(song_id)), "link": "detail"}
     )
-    return _parse_song_page(BeautifulSoup(html, "lxml"), song_id)
+    return _parse_song_page(BeautifulSoup(html, "lxml"), song_id, logger)
 
 
-def _parse_song_page(soup: BeautifulSoup, song_id: SongId) -> SongDetails:
+def _parse_song_page(
+    soup: BeautifulSoup, song_id: SongId, logger: Log = _logger
```
RumovZ, line 294: I think we should better address that in the testing code, e.g. by passing a fake Logger like `get_logger(__file__)`.

src/usdb_syncer/usdb_scraper.py
```diff
@@ -369,7 +371,10 @@ def get_usdb_available_songs(
 
 
 def _parse_details_table(
-    details_table: BeautifulSoup, song_id: SongId, usdb_strings: Type[UsdbStrings]
+    details_table: BeautifulSoup,
+    song_id: SongId,
+    usdb_strings: Type[UsdbStrings],
+    logger: Log,
```
RumovZ, line 377: Functions with a song id actually don't need to be passed a logger, as they can get it themselves: `get_logger(__file__, song_id)`.

RumovZ (approved):

PR #164, opened 2023-09-18
Initial steps towards a delete action for a song folder.

bohning:
This seems to be already working as intended - however it needs further improvements:

- [ ] extend by a confirmation dialog where the delete operation can also be cancelled
- [ ] update the song table to indicate that the local files are not available anymore
- [ ] the delete operation should work for all songs of a selection.


RumovZ:
Added in 936e35e6196c55f846665a5228a27d02b1a4f00b.

Issue #169, opened 2023-09-23
VIDEO tag is being parsed even though it's a local file


Baklap4:
Song: https://usdb.animux.de/?link=detail&id=20403
syncer: latest main
Reproduce:
1. Download the song
2. Restart the syncer
3. [WARNING] #20403: unknown key for meta tag: 'Alessio Bernabei - Io e te = la soluzione.mp4

This is because the parsed `VIDEO` tag in the locally stored txt file contains the `=` character this messes with the metatag extraction part

easy workaround: Change title to not contain the `=` as this is how the video tag gets filled i think

mjhalwa:
I could reproduce on windows on f1058aa .

I had a look to the code to find the exact location this is raised:

1. in [`cli_entry()` of `usdb_syncer/main.py`](https://github.com/bohning/usdb_syncer/blob/f1058aa9959e5b2333e7e8140d8fa1c9eb9c25d3/src/usdb_syncer/main.py#L29-L31) runs `main()` of `usdb_syncer/gui/mw.py`
2. in [`main()` of `usdb_syncer/gui/mw.py`](https://github.com/bohning/usdb_syncer/blob/f1058aa9959e5b2333e7e8140d8fa1c9eb9c25d3/src/usdb_syncer/gui/mw.py#L415) runs `_load_main_window(mw)` in same module
3. in [`_load_main_window(mw: MainWindow)` of `usdb_syncer/gui/mw.py`](https://github.com/bohning/usdb_syncer/blob/f1058aa9959e5b2333e7e8140d8fa1c9eb9c25d3/src/usdb_syncer/gui/mw.py#L424) runs `songs = get_all_song_data(False)` of `usdb_syncer/song_list_fetcher.py`
4. in [`def get_all_song_data(force_reload: bool)` of `usdb_syncer/song_list_fetcher.py`](https://github.com/bohning/usdb_syncer/blob/f1058aa9959e5b2333e7e8140d8fa1c9eb9c25d3/src/usdb_syncer/song_list_fetcher.py#L22) runs `local_files = find_local_files()` in same module
[...]

PR #170, opened 2023-09-23
avoid parsing meta-tags on updating local file list

mjhalwa:
As local files should not contain any meta-tags any more I propose fixing #169 by introducing a bool to skip parsing meta tags when loading local file list to (prior)

```
def find_local_files() -> dict[SongId, LocalFiles]:
    local_files: dict[SongId, LocalFiles] = {}
    for path in settings.get_song_dir().glob("**/*.usdb"):
        if meta := SyncMeta.try_from_file(path):
            local_files[meta.song_id] = files = LocalFiles(usdb_path=path)
            if txt := _get_song_txt(meta, path.parent):
                files.txt = True
                files.audio = _file_exists(path.parent, txt.headers.mp3)
                files.video = _file_exists(path.parent, txt.headers.video)
                files.cover = _file_exists(path.parent, txt.headers.cover)
                files.background = _file_exists(path.parent, txt.headers.background)
    return local_files
```

because the parsed meta-tags of `txt` are not used anywhere here.

But I am open for other suggestions ;)

### DIFF
--- a/src/usdb_syncer/song_list_fetcher.py
+++ b/src/usdb_syncer/song_list_fetcher.py
@@ -74,7 +74,9 @@ def find_local_files() -> dict[SongId, LocalFiles]:
     for path in settings.get_song_dir().glob("**/*.usdb"):
         if meta := SyncMeta.try_from_file(path):
             local_files[meta.song_id] = files = LocalFiles(usdb_path=path)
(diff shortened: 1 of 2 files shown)


mjhalwa:
rebased to changes in `main` branch

PR #174, opened 2023-09-29
Add a logger warning of #LANGUAGE tag is missing.

bohning:


RumovZ:
I think this should be logged on the debug level, like other messages of the same kind.

RumovZ (approved):

PR #175, opened 2023-09-30
Profiling

### LINKED ISSUE #156: Slow startup after downloading all the files
Heey folks,

I've been testing quite a bit with the new release and i love this project it helps me get the latest and greatest songs which is great!
Thanks for making it!

However starting the application (after downloading all of the songs) takes quite a while with lots of diskactivity.
I'm not quite sure how the loading is done, but i do see a `<songId>.usdb` file in every folder which is probably used for checking the songs during startup?

I'm ussually waiting for 10 to 20 minutes on my harddisk (which isn't throttling during loading) for the gui to popup.
This is quite the wait... this is probably due to having to go through each directory to get the info from these files?
If we can have one overlapping file within the `C:\Users\<user>\AppData\Local\bohning\usdb_syncer\Cache` folder this would immensly speed up the loading

As a side note; during startup a splashscreen is shown, however the splashscreen doesn't have a windows taskbar icon so it's only viewable by minimizing other programs

### LINKED ISSUE #169: VIDEO tag is being parsed even though it's a local file

RumovZ:
Fixes #169. Closes #170. Maybe we can also close #156 now. I assume, deferring the scan for local files wouldn't make much of difference anymore, but we should look at a flamegraph for a launch with all ~25k songs first.


bohning (approved):

bohning:
Phenomenal! Thanks a ton!

Baklap4:
In regard of https://github.com/bohning/usdb_syncer/pull/175/commits/37a755578ef54ec3be81b75b528b60af03da2b90 (the extra description, quoted below)

> Note: This doesn't work retroacively. Previously downloaded audio files
will have a wrong recorded mtime and thus will have to be downloaded
again before this fix takes effect.

Maybe this should be mentioned within the release page? @bohning @RumovZ

RumovZ:
I didn't want to get too technical, because most users probably won't notice anyway, but I wouldn't mind a brief note.

PR #183, opened 2023-10-09
Use a connectionpool with a limit of 100 connections to usdb.

Baklap4:
The default http-pool is 10 connections. To increase this we must create an adapter and then mount that to the session.

Current value is 100 concurrent connections.

This improves the calls done with the session. So not video's downloading through yt-dlp.
Haven't found a way to parallize that more


RumovZ:
In a quick and dirty test, this didn't seem to speed things up. I guess the bottleneck is the `QThreadPool.maxThreadCount` anyway. You could experiment with that and measure the difference it makes.

bohning:
Further investigation required, closing for now.

PR #189, opened 2023-10-20
Handle full fanart.tv urls in metatags.

bohning:


src/usdb_syncer/meta_tags.py
```diff
@@ -90,8 +90,15 @@ class ImageMetaTags:
     resize: ResizeMetaTags | None = None
     contrast: Literal["auto"] | float | None = None
 
-    def source_url(self) -> str:
+    def source_url(self, logger: Log) -> str:
         if "://" in self.source:
+            if "fanart.tv" in self.source:
+                logger.warning(
```
RumovZ, line 96:
```suggestion
                logger.debug(
```
Warnings should be reserved for events that affect the user. Also, it feels a bit odd to tell the user to use fanart ids, because they are generally not the one who entered the URL.

RumovZ (changes requested):

src/usdb_syncer/meta_tags.py
bohning: True, I will change it to debug level, the user it not affected. I get your second point, but on the other hand, everybody could help improve the metatags by submitting an edit to usdb, not only the creator of the metatags.

RumovZ (approved):

PR #190, opened 2023-10-20
Use pin instead of tick icon for pinned status.

bohning:


src/usdb_syncer/gui/song_table/table_model.py
```diff
@@ -234,3 +234,7 @@ def yes_no_str(yes: bool) -> str:
 @cache
 def optional_check_icon(yes: bool) -> QIcon | None:
     return QIcon(":/icons/tick.png") if yes else None
+
+
```
RumovZ, line 238:
```suggestion

@cache
```
Icon generation seems to be relatively expensive, so we should cache it.

RumovZ (changes requested):

RumovZ (approved):

PR #192, opened 2023-11-04
adds missing commands for a successful installation from source to readme

### LINKED ISSUE #191: Installation issue: ModuleNotFoundError: No module named 'usdb_syncer.gui.forms'
Hi,

building and running this app is very hard ... currently I have this problem

`
[07:39:19]> ./USDBSyncer
ERROR:root:Uncaught exception:
    Traceback (most recent call last):
      File "usdb_syncer/main.py", line 66, in <module>
    cli_entry(run_tools=False)
      File "usdb_syncer/main.py", line 56, in cli_entry
    from usdb_syncer.gui.mw import main  # pylint: disable=import-outside-toplevel
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "PyInstaller/loader/pyimod02_importers.py", line 391, in exec_module
      File "usdb_syncer/gui/mw.py", line 22, in <module>
    from usdb_syncer.gui.about_dialog import AboutDialog
      File "PyInstaller/loader/pyimod02_importers.py", line 391, in exec_module
      File "usdb_syncer/gui/about_dialog.py", line 7, in <module>
    from usdb_syncer.gui.forms.AboutDialog import Ui_Dialog
    ModuleNotFoundError: No module named 'usdb_syncer.gui.forms'
[185429] Failed to execute script 'main' due to unhandled exception!
`
can anyone help?

Thx in advance
Roger

mjhalwa:
As installation requires non-documented extra steps, wanted to add these steps for all platforms. Thus fixes #191 for future users that want to install from source

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,16 +126,25 @@ If you don’t want to run the whole test pipeline, you can also use single comm
 ### Building
 
 **USDB Syncer** uses [pyinstaller](https://github.com/pyinstaller/pyinstaller) for creating
-executables. In order to build **USDB Syncer**, run the following command
+executables. In order to build **USDB Syncer**, run the following command (inside `pipenv shell`)
 
 ```bash
 # On Linux
+export PYTHONPATH="$PYTHONPATH:./src"
+python -m src.tools.write_release_info -p src/usdb_syncer/constants.py -v $(git tag | tail -n 1) -c $(git log | head -n 1 | cut -d " " -f 2)
+python -m src.tools.generate_pyside_files
 pyinstaller --name "USDBSyncer" --onefile src/usdb_syncer/main.py
 
 # On macOS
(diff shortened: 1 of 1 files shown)


RumovZ:
I assume that user was just trying to run the app, and falsely assumed that he had to build it first. If he had run it as described in the README, the GUI files would have already been generated.

Thanks for updating the build steps, but maybe we should remove this part from the README entirely to avoid confusion? Most users won't have any need to build themselves, and we should actually discourage it in our own interest: Otherwise we could end up with multiple executables *with the same version* floating around, which could make tracking down bugs a nightmare. You can't even rely on the commit hash because there is no guarantee that the working tree had been clean at the time of building.

Also, arguably the best way to build is to tag and push to your own fork, anyway.

PR #200, opened 2023-12-29
Feature: export songlist in JSON format

mjhalwa:
adds export of songlist in JSON format, which may be used to visualize in a custom webpage like https://mindtakerr.github.io/ (repo: https://github.com/mindtakerr/ultrastar-songlist)

expected structure:

``` json
{
        "songs": {
            "id": int,
            "artist": string,
            "title": string,
            "year": int | null,
            "edition": string | null,
            "genre": string | null,
            "language": string,
            "golden_notes": bool,
            "cover-url": string | null,
            "cover-meta": string | null,
            "cover-preview": string (base64, 100x100) | null,
            "audio-url": string | null,
            "video-url": string | null,
          ,
        "date": "YYYY-MM-DD hh:mm:ss",
        "version": int,
    }
```

see attached json as an example export of 5 songs:

[2023-12-29_songlist.json](https://github.com/bohning/usdb_syncer/files/13790199/2023-12-29_songlist.json)

- [x] currently probably a little hacky, would appreciate ideas for better ways to collect all data **[EDIT]** most recent changes in January 2024 allowed better acces to txt file path
- [x] need to apply formatter and check linter outputs, just dropped current progress
- [x] currently incompatible to JSON imports **[EDIT]** changed JSON import format and corresponding unit tests

**[EDIT]** 
**!!!** Introduces **BREAKING CHANGE** to JSON import format **!!!**


mjhalwa:
changed to this format as dashes are sometimes problematic to parse, adds version of syncer that produced this exported file
```
{
    "songs": {
        "id": int,
        "artist": string,
        "title": string,
        "year": int | null,
        "edition": string | null,
        "genre": string | null,
        "language": string,
        "golden_notes": bool,
        "cover_url": string | null,
        "cover_meta": string | null,
        "cover_preview": string (base64, 100x100) | null,
        "audio_url": string | null,
        "video_url": string | null,
        },
    "date": "YYYY-MM-DD hh:mm:ss",
    "syncer_version": string,
}
```

mjhalwa:
rebased on #223 to fix following linting error in `pdf.py` of current `main` that is changed in #223 anyways

```
py311-lint: commands[3]> black . --check --diff
--- C:\Daten\Git\usdb_syncer\usdb_syncer-feat_json_export\src\usdb_syncer\pdf.py        2024-01-27 20:46:46.956117+00:00+++ C:\Daten\Git\usdb_syncer\usdb_syncer-feat_json_export\src\usdb_syncer\pdf.py        2024-01-27 21:27:36.103495+00:00@@ -18,18 +18,16 @@
     section1: dict[str, list[Any]] = {}
     document["sections"].append(section1)
     content1: list[Any] = []
     section1["content"] = content1
     date = datetime.datetime.now()
-    content1.append(
-        {
-            ".": f"Songlist ({date:%Y-%m-%d})",
-            "style": "title",
-            "label": "title1",
-            "outline": {"level": 1, "text": "A different title 1"},
-        }
-    )
+    content1.append({
+        ".": f"Songlist ({date:%Y-%m-%d})",
+        "style": "title",
+        "label": "title1",
+        "outline": {"level": 1, "text": "A different title 1"},
+    })

     for song_id in songs:
         song = UsdbSong.get(song_id)
         if song:
             data = f"{song.song_id}\t\t{song.artist}\t\t{song.title}\t\t{song.language}"
would reformat C:\Daten\Git\usdb_syncer\usdb_syncer-feat_json_export\src\usdb_syncer\pdf.py

Oh no! 💥 💔 💥
1 file would be reformatted, 60 files would be left unchanged.
py311-lint: exit 1 (1.48 seconds) C:\Daten\Git\usdb_syncer\usdb_syncer-feat_json_export> black . --check --diff pid=16940
py311-lint: FAIL ✖ in 1 minute 24.91 seconds
```

mjhalwa:
Seems to work as expected, but is blocked by #223, currently requiring to install NotoSans-Regular.ttf and NotoSans-Bold.ttf, else usdb_syncer crashes on start.

src/usdb_syncer/json.py
```diff
@@ -0,0 +1,79 @@
...
+                and meta.meta_tags.player2 is not None
+            ),
+        }
+        for song_id in songs
+        if (song := UsdbSong.get(song_id))
+        and (meta := song.sync_meta)
+        and (txt := meta.txt)
+        and (headers := get_headers(str(meta.path.with_name(txt.fname))))
+    ]
+    content = {"songs": song_list, "date": str(date), "syncer_version": VERSION}
```
RumovZ, line 74: The format should have its own versioning.

src/usdb_syncer/json.py
```diff
@@ -0,0 +1,79 @@
...
+
+_logger = get_logger(__file__)
+
+MAX_SIZE = (100, 100)
+
+
+def get_headers(txt_path: str) -> Headers:
+    with open(txt_path, "r", encoding="utf-8") as file:
+        lines = [line for line in file.read().splitlines() if line]
+        return Headers.parse(lines, _logger)
```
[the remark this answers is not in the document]
mjhalwa: `year`, `edition`, `genre` were not available internally in the syncer in my first atempt end of December - is it now? Thus I reparsed the txt and as I already did so also used the `artist` and `title` from the txt as well to get any user specific changes.

src/usdb_syncer/json.py
mjhalwa: I thought some information might be related on a syncer update and therefore written differently. This would lead to different output of the same song even though the actual json format did not change. Also I thought the JSON format will change less frequently than the syncer version so in case some compatibility issues arise we could mark ranges of syncer versions as compatible. Introducing an additional versioning seemed confusing to me.

src/usdb_syncer/json.py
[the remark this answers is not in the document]
RumovZ: All attributes from the USDB song list have always been available. This includes edition, but not year and genre. It would make sense to store these attributes somewhere after downloading a new txt.
But if we now start to dynamically read them from disk in some parts of the code, it will be harder to unify later on.
Most songs don't have these attributes anyway. Can't we just leave them out for now?

src/usdb_syncer/json.py
RumovZ: > I thought some information might be related on a syncer update and therefore written differently. [...]  we could mark ranges of syncer versions as compatible.

That seems to defeat the whole purpose of a format. Consumers would need to be aware of the producer and all its versions.

src/usdb_syncer/json.py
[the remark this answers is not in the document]
mjhalwa: IMHO they appear relevant for a song list export. Otherwise the ability to sort/filter songs for `genre` or `year` relies on an additional fetch from usdb.
I could try to extend `UsdbSong` accordingly so that it contains `genre` and `year`. We might also need to add `tags` once they are supported in one way or the other (not now)
There is already an if condition in place that handles possible inavailability of the values.

src/usdb_syncer/json.py
mjhalwa: I see, so I introduce a json export format version as local variable independent from the syncer version and start with `0.1.0`?

src/usdb_syncer/json.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def generate_song_json(songs: Iterable[SongId], path: str) -> int:
```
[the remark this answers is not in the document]
mjhalwa: done in 92c2c3b

I originally copied the definition from [`pdf.py`](https://github.com/bohning/usdb_syncer/blob/main/src/usdb_syncer/pdf.py#L13)
``` py
def generate_song_pdf(songs: Iterable[SongId], path: str) -> None:
```
and the function call is in [`mw.py](https://github.com/mjhalwa/usdb_syncer/blob/9381494daa7dfb1536299c7a91625e4498aaea6c/src/usdb_syncer/gui/mw.py#L187) where `path` is generated (equally as for the pdf generation) by
``` py
path = QFileDialog.getSaveFileName(self, dir=path, filter="JSON (*.json)")[0]
```
that responds a `str`.

If I changed to `Path` here, shouln't I change to `Path` in `pdf.py` as well?

src/usdb_syncer/json.py
```diff
@@ -0,0 +1,79 @@
...
+from usdb_syncer import SongId
+from usdb_syncer.constants import VERSION
+from usdb_syncer.logger import get_logger
+from usdb_syncer.resource_dl import _url_from_resource
+from usdb_syncer.song_txt.headers import Headers
+from usdb_syncer.usdb_scraper import UsdbSong
+
+_logger = get_logger(__file__)
+
+MAX_SIZE = (100, 100)
```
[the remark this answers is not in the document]
mjhalwa: done in 92c2c3b

src/usdb_syncer/json.py
```diff
@@ -0,0 +1,79 @@
...
+import datetime
+import json
+from typing import Any, Iterable
+
+from usdb_syncer import SongId
+from usdb_syncer.constants import VERSION
+from usdb_syncer.logger import get_logger
+from usdb_syncer.resource_dl import _url_from_resource
+from usdb_syncer.song_txt.headers import Headers
+from usdb_syncer.usdb_scraper import UsdbSong
```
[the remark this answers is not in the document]
mjhalwa: done in 92c2c3b

src/usdb_syncer/json.py
RumovZ: Yes, I would remove `syncer_version`, call the field `version` and make it a simple integer.

mjhalwa: added `export_version` in ce58e2c

src/usdb_syncer/json.py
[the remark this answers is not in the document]
RumovZ: I can add support for these attributes when I get a chance. You could set them to `None` for now, and we populate them as soon as the syncer supports it.

src/usdb_syncer/json.py
```diff
@@ -0,0 +1,79 @@
...
+                if (not headers.edition or headers.edition == "None")
+                else headers.edition
+            ),
+            "genre": (
+                None
+                if (not headers.genre or headers.genre == "None")
+                else headers.genre
+            ),
+            "language": song.language,
+            "golden_notes": bool(song.golden_notes),  # force true/false instead of 1/0
```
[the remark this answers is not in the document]
mjhalwa: fixed in 6596382, though unsatisfying to just put the `bool(row[5])` there, I hoped to find some db table column definition where I would expect golden_notes to be defined as `INT` that I could turn into `BOOLEAN` so that database already returns a proper data type, but couldn't find it.

src/usdb_syncer/json.py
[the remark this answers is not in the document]
mjhalwa: fixed the correlated linting error in `mw.py` in 2372e59

src/usdb_syncer/json.py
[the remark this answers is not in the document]
RumovZ: SQLite only supports data types TEXT, NUMERIC, INTEGER, REAL and BLOB.

src/usdb_syncer/json.py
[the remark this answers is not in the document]
RumovZ: I think we will remove pdf.py anyway.

src/usdb_syncer/json.py
```diff
@@ -0,0 +1,79 @@
...
+
+def get_headers(txt_path: str) -> Headers:
+    with open(txt_path, "r", encoding="utf-8") as file:
+        lines = [line for line in file.read().splitlines() if line]
+        return Headers.parse(lines, _logger)
+
+
+def generate_song_json(songs: Iterable[SongId], path: str) -> int:
+    date = datetime.datetime.now()
+    song_list: list[dict[str, Any]] = [
```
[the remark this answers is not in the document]
mjhalwa: Thanks for pointing me to that, didnt know about it. I changed as requested in fbeafba - I guess `date` could be some sort of default as well, but in `attrs.field(default=...)` and in `from_songs(..., date: datetime.datetime = ...)` the generated date was always the application startup date, so I ended up generating date in `generate_song_json()`

src/usdb_syncer/json.py
mjhalwa: just a single integer `version` in b46a14a

src/usdb_syncer/json.py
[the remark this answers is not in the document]
mjhalwa: done in 1a2db45

src/usdb_syncer/json.py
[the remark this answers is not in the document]
RumovZ: You're probably looking for `factory` kwarg.

tests/resources/import/ids.json
```diff
@@ -1,11 +1,13 @@
...
+      "id": "00001"
+    },
+    {
+      "id": 11586
+    },
+    {
+      "id": 3
+    }
+  ]
+}
```
RumovZ, line 13: These don't match the format you specified. In your comment and the Python implementation they are nullable, i.e. the key must exist, but the value may be null. Here, it seems as if they are optional, i.e. not even the key is required.

mjhalwa:
> rebased on #223 to fix following linting error in `pdf.py` of current `main` that is changed in #223 anyways
> 
> ```
> py311-lint: commands[3]> black . --check --diff
> --- C:\Daten\Git\usdb_syncer\usdb_syncer-feat_json_export\src\usdb_syncer\pdf.py        2024-01-27 20:46:46.956117+00:00+++ C:\Daten\Git\usdb_syncer\usdb_syncer-feat_json_export\src\usdb_syncer\pdf.py        2024-01-27 21:27:36.103495+00:00@@ -18,18 +18,16 @@
>      section1: dict[str, list[Any]] = {}
>      document["sections"].append(section1)
>      content1: list[Any] = []
>      section1["content"] = content1
>      date = datetime.datetime.now()
> -    content1.append(
> -        {
> -            ".": f"Songlist ({date:%Y-%m-%d})",
> -            "style": "title",
> -            "label": "title1",
> -            "outline": {"level": 1, "text": "A different title 1"},
> -        }
> -    )
> +    content1.append({
> +        ".": f"Songlist ({date:%Y-%m-%d})",
> +        "style": "title",
> +        "label": "title1",
> +        "outline": {"level": 1, "text": "A different title 1"},
> +    })
> 
>      for song_id in songs:
>          song = UsdbSong.get(song_id)
>          if song:
>              data = f"{song.song_id}\t\t{song.artist}\t\t{song.title}\t\t{song.language}"
> would reformat C:\Daten\Git\usdb_syncer\usdb_syncer-feat_json_export\src\usdb_syncer\pdf.py
> 
> Oh no! 💥 💔 💥
> 1 file would be reformatted, 60 files would be left unchanged.
> py311-lint: exit 1 (1.48 seconds) C:\Daten\Git\usdb_syncer\usdb_syncer-feat_json_export> black . --check --diff pid=16940
> py311-lint: FAIL ✖ in 1 minute 24.91 seconds
> ```

removed this PR again from history of this PR, no again just rebased to `main`

tests/resources/import/ids.json
mjhalwa: I just wanted to test the bare minimum as the import ignores all other fields. - should I use the exact export format here?

src/usdb_syncer/json.py
```diff
@@ -0,0 +1,114 @@
+"""Generates a JSON from the passed song list."""
```
RumovZ, line 1: Perhaps the module name should be changed to avoid name conflicts with the module from the stanard library.

src/usdb_syncer/json.py
```diff
@@ -0,0 +1,114 @@
...
+import json
+from json import JSONEncoder
+from pathlib import Path
+from typing import Any, Iterable
+
+import attrs
+
+from usdb_syncer import SongId
+from usdb_syncer.logger import get_logger
+from usdb_syncer.resource_dl import _url_from_resource
```
RumovZ, line 15: I'm surprised pylint doesn't complain about access of a private variable here.
But you can just mark it public by removing the leading underscore and potentially moving it to utils.py.

tests/resources/import/ids.json
mjhalwa: The purpose of the export format is to export all available information, the purpose of the import format is to define a couple of ids for import, where data and infos are retrieved from USDB. Within this PR it seemed weird if the import JSON format is totally incompatible with the export JSON Format. So I changed this that an exported JSON may be imported again, as well as self-generated JSON just from a couple of ids in proper way.

RumovZ: Would be odd if the fixtures didn't match the format. But it's your format, you can just decide that these field are optional. 🙂 Only then, they shouldn't also be nullable.

src/usdb_syncer/json.py
mjhalwa: oh true, this is still a dirty hack. Is there a public alternative - or should I make this function public?

src/usdb_syncer/json.py
mjhalwa: I think `json_export.py` makes sense or just `export.py` if we merge with `pdf.py`

src/usdb_syncer/json.py
mjhalwa: done in 61e57da

tests/resources/import/ids.json
mjhalwa: I don't see the point of making the JSON import format identical with the JSON export format, my goal was to make the export format compatible with the import format, not the other way round:

1. If this import format would not be valid, one that just wants to import a couple of usdb_ids from USDB would be force to use the `.usdb_ids` file or would have populate the JSON with extra fields that have no specific purpose and would not need to be correct for the import to work.
2. In addition if I added and checked the extra fields in the import format I would actually also need to check e.g. the `title` and `artist` to match USDB and inform the user if they changed or do some other action if I see changes.
3. Finally a user might not know the correct artist and title and just have a list of ids to import. The user could just wild guess titles and artists and the import would still work or fail if not guessed correctly.

Do we really want to validate the JSON format on import match the JSON format on export?

RumovZ: > I don't see the point of making the JSON import format identical with the JSON export format

I agree, there's no reason the ID JSON *had* to be the same as the song JSON. But since you've deliberately changed the fixtures, I thought that would be your plan.

src/usdb_syncer/json.py
RumovZ: Since pdf.py will be removed for now, json_export.py sounds good. Also export.py is all fun and games until you want to add the import equivalent. 😄

RumovZ (approved):

Issue #20, opened 2022-11-15
Add basic cleanup tasks for songs


bohning:
The tool could perform some basic cleanup/fixes on the songs, e.g.

- shift first note to beat 0 (and adjust #GAP accordingly)
- normalize pitch values (sometimes they are in the 60s --> towards a zero mean)
- (repeatedly) double BPM and all note timings if BPM is very low
- shorten notes that touch (if duration > 1)

This should be done for the txt file that is generated from the .usdb file, not on the .usdb file itself, so these changes are local.

These could be made selectable by the user (who can choose which fixes he/she wants), but the downside of that is that different users have different variants of the song file.

DeinAlptraum:
> replace upright apostrophe ' by typographer’s apostrophe ’ (optional?)
I would really appreciate if this could be made optional (i.e. allowing to choose the apostrophe you prefer)

bohning:
> I would really appreciate if this could be made optional (i.e. allowing to choose the apostrophe you prefer)
What would you choose and why?

DeinAlptraum:
I prefer the straight apostrophe ' visually.
I've also had various problems in the past with files with typographer's apostrophe in their name not being recognized by USDX, though it's been a while, but I would prefer to eliminate this source of errors to be sure.

bohning:
I guess the best option to handle notes of zero length is to extend them to 1 beat duration if there is 2 beats space to the beginning of the next note. What do you think?

bohning:
How shall we handle txt files without linebreak timings? The syncer could insert linebreak timings (then they would work with the current USDX), or it could consider the linebreak timing optional. Currently, the linebreaks are deleted entirely by the syncer, rendering the file unusable (related: USDX refuses to load files that do not contain a single line break, presumably because songs with a single line of lyrics are rather uncommon, but exceptions exist: https://usdb.animux.de/index.php?link=detail&id=25978).

samcarton:
> > I would really appreciate if this could be made optional (i.e. allowing to choose the apostrophe you prefer)
> 
> 
> 
> What would you choose and why?
> 
> 

I'm facing issues on steam deck (arch Linux), USDX doesn't pick up any songs with the typographic apostrophe. If I manually change them back to the typewriter apostrophe (0027) then it has no problem.
Ideally usdb syncer would handle this for me otherwise I need to script it up myself.

bohning:
[...]

PR #202, opened 2024-01-04
Add an option to choose the preferred apostrophe

DeinAlptraum:
Add an option to choose between typographers apostrophe (default) and upright apostrophe, to be used when sanitizing file paths and song artist, title and lyrics, as I requested on #20 

Changing the value has no immediate affect, only when a file is re-downloaded, it changes the folder/file names and the content of the txt. This also works when the file hasn't been changed on USDB side.

### DIFF
--- a/src/usdb_syncer/download_options.py
+++ b/src/usdb_syncer/download_options.py
@@ -7,6 +7,13 @@ from usdb_syncer import settings
 
 
 @dataclass(frozen=True)
+class MiscOptions:
+    """Miscellaneous settings that must always be present."""
+
+    apostrophe: settings.Apostrophe
+
+
+@dataclass(frozen=True)
 class TxtOptions:
     """Settings regarding the song txt file to be downloaded."""
 
@@ -67,6 +74,7 @@ class Options:
     """Settings for downloading songs."""
 
     song_dir: Path
+    misc_options: MiscOptions
     txt_options: TxtOptions | None
     audio_options: AudioOptions | None
     browser: settings.Browser
@@ -78,6 +86,7 @@ class Options:
 def download_options() -> Options:
     return Options(
         song_dir=settings.get_song_dir(),
+        misc_options=_misc_options(),
         txt_options=_txt_options(),
         audio_options=_audio_options(),
         browser=settings.get_browser(),
@@ -85,6 +94,10 @@ def download_options() -> Options:
         cover=_cover_options(),
         background_options=_background_options(),
     )
+
+
+def _misc_options() -> MiscOptions:
+    return MiscOptions(settings.get_apostrophe())
 
 
 def _txt_options() -> TxtOptions | None:
--- a/src/usdb_syncer/gui/settings_dialog.py
+++ b/src/usdb_syncer/gui/settings_dialog.py
@@ -24,6 +24,8 @@ class SettingsDialog(Ui_Dialog, QDialog):
             self.comboBox_encoding.addItem(str(encoding), encoding)
         for newline in settings.Newline:
             self.comboBox_line_endings.addItem(str(newline), newline)
+        for apostrophe in settings.Apostrophe:
+            self.comboBox_apostrophes.addItem(str(apostrophe), apostrophe)
         for container in settings.AudioFormat:
             self.comboBox_audio_format.addItem(str(container), container)
         for bitrate in settings.AudioBitrate:
@@ -51,6 +53,9 @@ class SettingsDialog(Ui_Dialog, QDialog):
         )
         self.comboBox_line_endings.setCurrentIndex(
             self.comboBox_line_endings.findData(settings.get_newline())
+        )
+        self.comboBox_apostrophes.setCurrentIndex(
+            self.comboBox_apostrophes.findData(settings.get_apostrophe())
         )
         self.groupBox_audio.setChecked(settings.get_audio())
         self.comboBox_audio_format.setCurrentIndex(
@@ -91,6 +96,7 @@ class SettingsDialog(Ui_Dialog, QDialog):
         settings.set_txt(self.groupBox_songfile.isChecked())
         settings.set_encoding(self.comboBox_encoding.currentData())
         settings.set_newline(self.comboBox_line_endings.currentData())
+        settings.set_apostrophe(self.comboBox_apostrophes.currentData())
         settings.set_audio(self.groupBox_audio.isChecked())
         settings.set_audio_format(self.comboBox_audio_format.currentData())
         settings.set_audio_bitrate(self.comboBox_audio_bitrate.currentData())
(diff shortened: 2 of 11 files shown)


src/usdb_syncer/download_options.py
```diff
@@ -6,6 +6,13 @@
 from usdb_syncer import settings
 
 
+@dataclass(frozen=True)
+class MiscOptions:
```
DeinAlptraum, line 10: Unfortunately this cannot be added to `TxtOptions` since this might not be presented when "Download Song file" is unchecked. This option always has to be present since it's also used to sanitize the paths made up of artist & title.
I didn't have a good idea for a class name.

src/usdb_syncer/gui/settings_dialog.py
```diff
@@ -24,6 +24,8 @@ def _populate_comboboxes(self) -> None:
             self.comboBox_encoding.addItem(str(encoding), encoding)
         for newline in settings.Newline:
             self.comboBox_line_endings.addItem(str(newline), newline)
+        for apostrophe in settings.Apostrophe:
+            self.comboBox_apostrophes.addItem(str(apostrophe), apostrophe)
```
DeinAlptraum, line 28: This has apparently been the final straw for `pylint` to rate the `_populate_comboboxes` function as "too complex".
I personally don't think this is a useful measure and should be deactivated. Alternatively, one could put the combobox names and setting enums into lists and have an outer for loop that iterates over these, but this isn't much prettier imo

bohning:
Sorry for not responding earlier. I truly appreciate your efforts, but I have to be honest here, as my heart not only beats for singing/karaoke, but also for typography.

Referring to the article about [apostrophes](https://de.wikipedia.org/wiki/Apostroph#Digitale_Typografie) on the German Wikipedia, the upright apostrophe is a typographically *wrong* character and stems from the old 7-bit ASCII character sets where symbol count was very limited. The upright apostrophe is solely a *substitute* character *when the correct apostrophe is not available*. Since we are moving away from ASCII/CP1252 or any other codepage encoding towards UTF8, we should embrace the versatility and the typographic possibilities this enables. supporting songs of any language/script (without needing to resort to romanization), we can support artists like "A★Teens" or titles like Prince’s "Sign “☮” The Times".

I *suspect* you dislike the typographer’s apostrophe because you have been used to the *substitute* character for so long. That reminds me of my father, who grew up using old typewriters that only had one key for O and 0, so still today he sometimes types "hell0" instead of "hello", just because he has gotten so used to it.

I agree that in the past, the typographer’s apostrophe sometimes lead to headaches, because it is not part of ASCII (but of CP1252, Pos. 92) and if the wrong encoding was used, it screwed up the text display. But with the support of UTF8, this is or should be a thing of the past. The same goes for using the typographer’s apostrophe - all somewhat modern operating systems support UTF8 filenames, so unless you’re using Windows 95, the typographer’s apostrophe in file- and folder names should not lead to trouble, neither from the OS side nor from the karaoke software side.

So to sum up, I truly believe we should aim for the *correct* way of doing it, and not continue using a crutch that dates back to 7-bit character encodings (1960s). Search queries, both in-game and on usdb, should handle the correct and the substitute apostrophe transparently, and at least in USDX, this was something you were already working on.

Hopefully, I was able to make it a little relatable why these decisions were made and why I am not in favor of accepting this PR, even though I truly appreciate the efforts.

Are there any pain points with the typographer’s apostrophe besides they way the apostrophe *looks*? Is anything not *working* that we should be aware of and that we haven’t taken into consideration?

irgendwr:
@bohning one important difference to your analogy (O vs 0) is that there is no a simple/universal key for the typographer's apostrophe. It might sound like a negligible pain point, but it's the reason why I still continue to use the upright apostrophe (as you can see here). The method for typing `’` varies depending on OS/software, and while one can look it up, it still takes more effort. When I write more formal/important texts, I do tend to put in the tiny extra effort and use the correct typographic symbols, but for "everyday activities" I choose convenience over being 100% correct.

Would you also want to force the use of typographically correct symbols in programming languages? (genuine question)

In my case you could probably just call it laziness or resistance to change, but in other cases it's a matter of accessibility.

Also, I just tried typing the typographer's apostrophe in UltraStar Deluxe on Linux and it doesn't work for me. How am I supposed to search for songs with an apostrophe then?

I definitely understand where you're coming from, but at the end of the day I think it should be up the the user to decide. So setting the typographically correct symbols as the default is reasonable, but I'd argue that it should still be a choice.

DeinAlptraum:
Thank you for your detailed answer!

@irgendwr has covered most of what I wanted to say already. While I understand your argument about correctness, and just being more used to the upright apostrophe etc., I believe this is not something that will ever change. The typographer's apostrophe is simply impractical to type, which is the main reason why comparatively few people are using it at the moment, and are still (and probably always will be) used to the upright apostrophe.

To me, this is mostly about customizability. Even though the upright apostrophe might be wrong, it is the most natural option for, I believe, the majority of people. As such I think it is reasonable to support this as an option in the syncer. This is a choice anyone can make for themselves, or is there any way this might affect other users that I am overlooking right now?

Regarding your question about pain points: I had issues with songs not being recognized by USDX in the past if they had typographer's apostrophes in the title/artist name, but the same was true for most non-ASCII characters... I solved this by fixing my broken locale settings (iirc) so might not have any broader relevance. Other than that and the WIP search normalization in USDX, no problems from my side.

bohning:
Thanks for your responses and the discussion, I really like it.

I understand the raised issue of accessibility (typing the typographer’s apostrophe *is* a pain in the butt for most people, no question about it), BUT: I am *not* asking anyone to type it.

Regarding file and folder names - unless I hear from someone otherwise, I expect all major operating systems to support UTF8 filenames. Thus this should not be an issue for anyone.

Regarding search queries:
The search in USDB Syncer handles either apostrophe in a transparent way, you can search for both "Don’t Cry" (typographer’s apostrophe) or "Don't cry" (upright apostrophe) and it will find songs with either apostrophe.

The search in USDB needs to be adopted to behave the same way *for those not using the Syncer*.

The search in USDX (and all the other karaoke software) needs be adapted to behave the same way. This will *automatically be solved when the transliteration of the search is in place*, and this is something that is needed anyhow when UTF8 is supported (the vast majority of UTF8 characters cannot be typed on a regular keyboard!).

Ideally (IMHO), all song creation tools and lyrics websites would already correct upright to typographer’s apostrophes so that you never have to type it manually (spoiler: they don’t). But even if you upload a song to usdb with upright apostrophes, that’s fine, since the Syncer will fix the apostrophes (amongst other things) automatically when downloading.

To summarize, the whole idea is to have the correct typographer’s apostrophes without having to worry a single bit about them. They just work, both on file/folder level and in all softwares. And when you need to type apostrophes, use whichever you can type easier (which is of course the upright apostrophe, at least for everyone else but me - I changed my keyboard layout so that SHIFT+# types the ’ and not ', that’s why I needed [PR791](https://github.com/UltraStar-Deluxe/USDX/pull/791)).

PR #216, opened 2024-01-22
Fix synchronize_sync_meta for macOS

bohning:
Ignore hidden "dot files" that start with "._".


RumovZ:
This excludes any files starting with `.` *or* `_` if I'm not mistaken, and we need to match `_`. But ignoring any files starting with `.` should be fine.

bohning:
> This excludes any files starting with `.` _or_ `_` if I'm not mistaken, and we need to match `_`. But ignoring any files starting with `.` should be fine.

Yes, I think you are right, good catch. I changed it accordingly to exclude all files starting with a dot.

RumovZ (approved):

PR #219, opened 2024-01-27
Db cleanup and other fixes

RumovZ:
Even without extra indices, searching by prefix doesn't feel really slower, and prefix indices are in fact doubling the db size. However, since we're still only talking a few MB here, I guess nobody will care and still added them in.

Side note: This affects db creation and requires deleting the old db file.


bohning (approved):
🚀🥳🙏

PR #222, opened 2024-01-27
Add support for opus audio codec.

### LINKED ISSUE #214: Add Opus sound format to grab from Youtube
Hi,

I'd like to suggest support for Opus. It's a free sound format that Youtube directly supports and should be pretty transparent at Youtube bitrates (better than ogg vorbis).

It is at least supported by USDX and Vocaluxe nightly.

bohning:
Closes #214. Closes #217.

### DIFF
--- a/src/usdb_syncer/song_loader.py
+++ b/src/usdb_syncer/song_loader.py
@@ -348,7 +348,7 @@ class _SongLoader(QtCore.QRunnable):
 
             # last chance to abort before irreversible changes
             self._check_flags()
-            _cleanup_existing_resource(ctx)
+            _cleanup_existing_resources(ctx)
             _ensure_correct_folder_name(ctx.locations)
             # only here so filenames in header are up-to-date
             _maybe_write_txt(ctx)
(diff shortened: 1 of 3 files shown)


src/usdb_syncer/song_loader.py
```diff
@@ -577,8 +579,15 @@ def _write_mp3_tags(
...
-    audio = mutagen.oggvorbis.OggVorbis(path)
+def _write_ogg_tags(
+    path: Path, resource: str, ctx: _Context, embed_artwork: bool
+) -> None:
+    audio: mutagen.ogg.OggFileType
+    match path.suffix:
+        case ".ogg":
+            audio = mutagen.oggvorbis.OggVorbis(path)
+        case ".opus":
+            audio = mutagen.oggopus.OggOpus(path)
```
RumovZ, line 590: If `path.suffix` is anything else, `audio` will be unbound and the function will throw. This is problematic because its signature doesn't state this requirement. pylint only accepts it due to the explicit annotation in line 585.
You could change the `path: Path` parameter to `audio: mutagen.ogg.OggFileType` and handle the ogg/opus distinction in the match statement above.

RumovZ (changes requested):

src/usdb_syncer/song_loader.py
bohning: Thanks, good point! You mean like this?

RumovZ: Yes, exactly. And thanks for correcting my typo. Should probably also be `_cleanup_existing_resourceS`, but whatever. 😅

RumovZ (approved):

PR #223, opened 2024-01-27
Fix pdf creation via using reportlab module.

bohning:
This fixes the PDF creation. However, this is still needs more work for it to be useful. It requires NotoSans font (https://fonts.google.com/noto/specimen/Noto+Sans) installed in order to display all Unicode characters. Not sure if we can and should ship the font files with the app, seems to be permissible by the font license.


RumovZ:
Deferred.

PR #227, opened 2024-02-12
Turn spinbox into combobox for max. cover size.

bohning:


src/usdb_syncer/settings.py
```diff
@@ -122,6 +122,34 @@ def default() -> Newline:
...
 
+class CoverMaxSize(Enum):
+    """Maximum cover size."""
+
+    PX_3840 = 3840
+    PX_1920 = 1920
+    PX_1200 = 1200
+    PX_1000 = 1000
+    PX_640 = 640
+    PX_500 = 500
```
RumovZ, line 133: Didn't you say 3 options would be enough? What changed your mind?
Also, it was possible to disable scaling completely before. Might be a viable alternative to the 3840px option.

src/usdb_syncer/gui/settings_dialog.py
```diff
@@ -19,11 +19,13 @@ def __init__(self, parent: QWidget) -> None:
         self._browser = self.comboBox_browser.currentData()
         self.groupBox_reencode_video.setVisible(False)
 
-    def _populate_comboboxes(self) -> None:
+    def _populate_comboboxes(self) -> None:  # pylint: disable=too-complex
```
RumovZ, line 22: We should address lints instead of disabling them. Otherwise, there's no value in using Pylint at all.
In this case, you should be able to build a tuple of all the enums and comboboxes and iterate over it. (Browser will need to be special-cased due to the icon.)

src/usdb_syncer/gui/forms/SettingsDialog.ui
```diff
@@ -124,35 +124,16 @@
...
-        <item row="0" column="1">
-         <widget class="QSpinBox" name="cover_max_size">
-          <property name="sizePolicy">
-           <sizepolicy hsizetype="Minimum" vsizetype="Fixed">
-            <horstretch>0</horstretch>
-            <verstretch>0</verstretch>
-           </sizepolicy>
-          </property>
-          <property name="toolTip">
-           <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;Images larger than this will be scaled to the set value. Use 0 to disable.&lt;/p&gt;&lt;p&gt;Some USDX variants have problems displaying covers larger than 1920x1920 pixels.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
```
RumovZ, line 136: This removed the tooltip.

RumovZ (changes requested):

src/usdb_syncer/settings.py
bohning: My reasoning was this:
1920x1920 is the max cover size for USDX on 4K screens. Consequently, 3940x3940 is the maximum needed by an 8K screen. 1200x1200 is the standard cover size provided by Apple, 1000x1000 is the standard cover size of fanart.tv, 640x640 is the standard cover size provided by spotify. 500x500 is for people who really try to save some disk space.

I agree that maybe we can change the 3940x3940 case to a disable option. What other option(s) would you get rid of?

src/usdb_syncer/gui/forms/SettingsDialog.ui
bohning: Good catch, that was not removed on purpose.

src/usdb_syncer/settings.py
RumovZ: I guess 1200 and either 500 or 640. Fanarts sounds like a reasonable benchmark, but the other ones probably aren't very relevant for the syncer, are they? Especially since users won't know the origin of these options anyway.

src/usdb_syncer/gui/settings_dialog.py
```diff
@@ -19,27 +19,28 @@ def __init__(self, parent: QWidget) -> None:
...
+        ]
+
+        for combobox, setting in combobox_settings:
+            for item in setting:
+                if combobox == self.comboBox_browser:
+                    combobox.addItem(
+                        (QIcon(settings.Browser(item).icon()), str(item), item)
+                    )
+                else:
+                    combobox.addItem((str(item), item))
```
RumovZ, line 43:
```suggestion
        combobox_settings = (
            (self.comboBox_encoding, settings.Encoding),
            (self.comboBox_line_endings, settings.Newline),
            (self.comboBox_cover_max_size, settings.CoverMaxSize),
            (self.comboBox_audio_format, settings.AudioFormat),
            (self.comboBox_audio_bitrate, settings.AudioBitrate),
            (self.comboBox_videocontainer, settings.VideoContainer),
            (self.comboBox_videoencoder, settings.VideoCodec),
            (self.comboBox_videoresolution, settings.VideoResolution),
            (self.comboBox_fps, settings.VideoFps),
        )
        for combobox, setting in combobox_settings:
            for item in setting:
                combobox.addItem((str(item), item))
        for browser in settings.Browser:
            self.comboBox_browser.addItem(QIcon(browser.icon()), str(browser), browser)
```

RumovZ (approved):

PR #240, opened 2024-03-15
Use a random instead of a hard-coded User-Agent for image requests.

bohning:


RumovZ:
Can we maybe solve the issue by fixing the hard-cded user agent? I'm afraid that if we use a random one, it might be impossible to troubleshoot reported download problems in the future.

Which command did you add the dependency with? It seems like pefile got dropped which appears to be required by pyinstaller on Windows.

RumovZ (approved):

PR #247, opened 2024-06-16
fix: Use `os.walk` instead of `pathlib.glob` for iterating usdb files

### LINKED ISSUE #244: Indexing existing song directory hangs
**Platform**

macOS 14.4.1 (Darwin)
Python 3.11.9
External SSD (USB 3.2), exFAT formatted
`usdb_syncer` version: 0.5.0 version from GitHub as well as the latest dev version (locally started in venv, commit: 5053d29

**Steps to reproduce**

- Download the entire Database to an external SSD with `usdb_syncer`
- Reset `usdb_syncer` (Settings, ...)
- Open `usdb_syncer` again and select song directory on external drive
- Watch `usdb_syncer` starting to index the song directory, after approx. 15 seconds no more console  output, the process is stuck (I let it run for 12 hours to ensure that this is not a display bug)

**Additional information**

I did the same on my Linux machine, there everything works fine.

g3n35i5:
For some reason, `pathlib.glob` hangs when iterating a large number of
files. When using `os.walk`, this problem seems to be fixed.

This resolve #244

### DIFF
--- a/src/usdb_syncer/song_routines.py
+++ b/src/usdb_syncer/song_routines.py
@@ -2,7 +2,9 @@
 
 import json
 import logging
+import os
 from pathlib import Path
+from typing import Generator
 
 from requests import Session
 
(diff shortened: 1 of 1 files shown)


RumovZ (approved):
I can't reproduce the issue, but the new code seems to be about 25% faster on my system.

bohning (approved):

PR #249, opened 2024-06-21
Moved meta files and special characters

### LINKED ISSUE #245: Problems with directory names, special characters and exFAT?
With my collection on an exFAT formatted SDXC card, I have the problem that redownloading a song that contains special characters leads to a redownload into a new folder instead of using the existing folder (and skipping the song if the resources haven't changed).

Example:
`Artist - Title with special character` directory already exists. Upon redownload, a second directory `Artist - Title with special character (1)` is created and the entire song is re-downloaded (so I end up having it twice). Thus, skipping existing songs for unchanged resources fails.

Not sure if this is Mac related or exFAT related or Mac+exFAT related, so would be interested if anyone can reproduce this (on Mac and on other platforms using an exFAT formatted external device).

I assume this has to do with Unicode normalization of directory names, googled a bit about it, but couldn't find anything specific to exFAT, but maybe it's related to Apple's HFS.

https://community.roonlabs.com/t/rock-having-problems-with-special-characters-german-aou-in-folder-names/119702
https://handbook.datalad.org/en/0.17/intro/filenaming.html
https://en.wikipedia.org/wiki/Filename
https://superuser.com/questions/999232/unicode-filenames-in-windows-vs-mac-os-x
https://stackoverflow.com/questions/74960030/rename-file-impossible-because-of-special-characters-in-filename

From the 4th link (superuser.com), one of the commenters states:

> **Windows uses UTF-16**. Most codepoints are encoded in two bytes. Codepoints outside the BMP (Basic Multilingual Plane) are encoding in a "surrogate pair" that takes four bytes. **Windows does not normalize filenames using any Unicode Normalization Form**.

> This means you could have two filenames that look identical with one using a precomposed "é" consisting of a single codepoint and the other using a regular ASCII "e" followed by a Unicode combining acute accent, thus two codepoints.

> **OS X uses UTF-8**. Codepoints are encoded using between one and five bytes. **OS X uses Unicode NFD (Normalization Form Canonical Decomposition)**.

> This means that when a Unicode character such as "é" is used in a filename it will always be normalized by the system into a regular ASCII "e" followed by a Unicode combining acute accent, and will always take two codepoints.

> In fact OS X uses the spec of Unicode NFD from either Unicode version 2.1 or 3.2, depending on the version of OS X.

The comment includes a link to Technical Note TN1150 about HFS details that is only available through the [wayback machine](https://web.archive.org/web/20200102075734/http://dubeiko.com/development/FileSystems/HFSPLUS/tn1150.html#UnicodeSubtleties)

### LINKED ISSUE #246: Problems with renaming directory names and exFAT
With my collection on an exFAT formatted SDXC, when I redownload a song where the Artist or Title has changed (thus leading to a renaming of the directory), the Syncer shows "Failed" in the Status column, but indeed it seems that everything worked just fine (directory has new name, all resources have new name).

RumovZ:
Fixes #245. Fixes #246.
@bohning, please let me know if it works on your system.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,14 @@
+<!-- 0.5.1 -->
+
+# Changes
+
+## Fixes
+
+- Fixed the app hanging on startup for some users with large collections on macOS.
+- Added detection and handling of moved meta files on startup or song folder change.
+- Downloading artist or title changes for existing songs no longer causes an error.
+- Fixed special characters causing issues on macOS in certain cases.
+
 <!-- 0.5.0 -->
 
 # Changes
--- a/src/usdb_syncer/song_loader.py
+++ b/src/usdb_syncer/song_loader.py
@@ -32,6 +32,7 @@ from usdb_syncer import (
     events,
     resource_dl,
     usdb_scraper,
+    utils,
 )
 from usdb_syncer.constants import ISO_639_2B_LANGUAGE_CODES
 from usdb_syncer.logger import Log, get_logger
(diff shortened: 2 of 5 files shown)


bohning:
Wow, works like a charm and indeed fixes #245 and #246. Simply awesome! Thanks a ton for fixing these annoying issues!

bohning (approved):

PR #254, opened 2024-06-28
Adapt to recent changes in usdb song list.

### LINKED ISSUE #253: Executable refuses to open with: "ValueError: invalid literal for int() with base 10: 'English'"
Since todays morning, the previously opening fine executable refused to open with the following log output:

image

I am happy to fix it myself if I am doing sth. wrong, but as I did not change something and this happens now, I thought I should open an Issue here.

bohning:
The song list returned by a search on usdb now contains columns for genre, year and creator.

What is still missing is the update of the song table to show the newly available information. Maybe you can help me with this, @RumovZ ?

Once merged, this closes #253 .


bohning:
Now genre and year were also added to the details page, so the parsing needs to be adapted to that as well.

RumovZ:
> What is still missing is the update of the song table to show the newly available information.

Already works great for me! Of course, the new data will only be available for entries newly fetched from USDB and not already in the local song list.

Refetching the song list failed due to a missing validity check, though.

I think the only thing left to do is adding search filters for the new values. Do you want to give it a try?

bohning:
The details page has changed as well (now includes year and genre), so the parsing of the details page still has to be adapted. I am on a trip at the moment, so feel free to finish this PR if you want to and have the time.

RumovZ:
I added the search filters.

The details page seems to parse fine, and we don't have any use for one of the new fields there, I think. But feel free to add them for consistency's sake.

RumovZ (approved):

PR #260, opened 2024-07-09
Hit space to preview/play the song's audio.

bohning:
If available, the syncer will play the local audio file, otherwise it will try to stream the audio sample from usdb if it is set.


RumovZ:
You probably have to reconfigure the required checks before this can be merged. I assume there are two checks, `tox` and `build`, and you have to remove `build`.

RumovZ (approved):

PR #269, opened 2024-07-20
Fix vimeo downloads by adding curl-cffi to yt-dlp.

bohning:
See https://github.com/yt-dlp/yt-dlp/issues/9682.


RumovZ (changes requested):
You have to either manually restore the removed dependenceis for other platforms by hand, or revert and use `pipenv update` to only modify yt-dlp.

RumovZ (approved):

PR #272, opened 2024-07-30
Improve FFmpeg dialog.

bohning:


bohning:
Good idea. I thought about that but didn't know exactly how.

RumovZ:
You could add one label per OS, then in the Python class disable the ones not matching the OS.

RumovZ (approved):

PR #273, opened 2024-08-09
Retrieve cookies from browser to access age restricted resources.

### LINKED ISSUE #248: Cookies seem to be not properly passed to yt-dlp
```
if path := browser.cookie_path():
        options["cookies"] = (f"{browser.value}:{path}",)
```
doesn't seem to do the trick, age restricted content still fails even though I'm logged into YT in a browser (test with e.g. https://usdb.animux.de/?link=detail&id=9743)

bohning:
This closes #248.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 ## Fixes
 
+- Age-restricted content can now be successfully downloaded if you are logged in to
+  YouTube in your browser.
+
 ## Features
 
 - The download path can now be customized using a dedicated template syntax (see
--- a/src/usdb_syncer/resource_dl.py
+++ b/src/usdb_syncer/resource_dl.py
@@ -126,9 +126,8 @@ def _ytdl_options(format_: str, browser: Browser, target_stem: Path) -> YtdlOpti
         # suppresses download of playlists, channels and search results
         "playlistend": 0,
         "overwrites": True,
+        "cookiesfrombrowser": (browser.value, None, None, None),
     }
-    if path := browser.cookie_path():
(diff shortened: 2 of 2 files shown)


Baklap4:
Doing this makes it so that the cookies get searched, loaded, extracted for each individual request, lots of I/O happening. Is there a way to optimize cookie usage?

It does fix problems related to ytdlp about not being logged in

bohning:
See https://github.com/yt-dlp/yt-dlp/issues/7973.

Baklap4:
Hmmm fair enough, then one more question, can we filter for the usdb and youtube cookie so we don't have to go through the entire list of cookies everytime?

Apparently i have around 3.5k cookies set and it takes a while to load and extract them all and using only 2

Baklap4:
Might be worth mentioning that if you download a whole bunch of files - being signed in (or rather use the browser-cookie) helps to prevent the "Are you a bot - please sign in" error message i've noticed. Not sure if you guys noticed this?

bohning:
I guess we need to decide if we want to
a) retrieve and use cookies for all requests in order to 1) avoid bot/sign-in issue and 2) to be on the safe side for age-restricted content - at the cost of the extra time needed to retrieve cookies or
b) generally use no cookies, unless age-restricted content is encountered
c) periodically use cookies (in order to avoid bot/sign-in issue) *and* when age-restricted content is encountered

Baklap4:
Hmmm i guess we can capture the 'are you not a bot please login' errormessage just like were doing with the age restricting content and retry it then?

Else option c sounds great to me

RumovZ (approved):

PR #276, opened 2024-08-13
Allow users to run custom commands after download

randompersona1:
This PR adds the ability to run a user-set command after the song has downloaded. This enables the syncer to be infinitely customizable at a users whim, for example, being able to split songs with AI or transcode videos to a more efficient codec.

settings screenshot

# Usage
A new setting with a plain text field allows entering the command. In the user-set command, occurrences of "$dir$" (writing this, I am not convinced of that name) will be replaced by the song folder that was just downloaded. The command is executed in the global song directory. A user could create a batch file there with the song folder as an argument, and write everything else in the batch file.

# Example Usage
I want to split my songs with [demucs](https://github.com/facebookresearch/demucs). Unfortunately, I could not get it to work with cuda in python3.12. The only thing I managed was to pin the torch dependencies to specific releases and use python3.11 for the project. This makes running that from python3.12 a pain. With this feature, I can write a little batch script that just passes the song folder to my program and launches it with poetry correctly (i could also enter the poetry command in the settings), therefore bypassing many requirements I assume are being written in the instrumental-version branch, for example actually installing the cuda drivers, and using the correct torch dependancy, because the cuda version of torch obviously doesn't work for users that don't have an nvidia graphics card. This feature allows users like me to take some of that responsibility, which definitely doesn't work for everyone, but greatly improves the experience those that can take advantage of it. 

# Implementation details
I decided not to write stdout of the command to the log, because that seemed like a great idea for completely spamming the log. The return code of the command is written to the log as an info, so there is that for communication. Otherwise, I've tried to include as much information about execution into the log, so there shouldn't be any confusion there. 

Regarding security, I don't think there is any need to consider it. We are only running a user-supplied command, at the same execution level as the rest of the syncer.

### DIFF
--- a/src/usdb_syncer/download_options.py
+++ b/src/usdb_syncer/download_options.py
@@ -63,6 +63,13 @@ class BackgroundOptions:
 
 
 @dataclass(frozen=True)
+class CommandOptions:
+    """Settings regarding the command to be executed."""
+
+    command: settings.Command
+
+
+@dataclass(frozen=True)
 class Options:
     """Settings for downloading songs."""
 
@@ -74,6 +81,7 @@ class Options:
     video_options: VideoOptions | None
     cover: CoverOptions | None
     background_options: BackgroundOptions | None
+    command_options: CommandOptions | None
 
 
 def download_options() -> Options:
@@ -86,6 +94,7 @@ def download_options() -> Options:
         video_options=_video_options(),
         cover=_cover_options(),
         background_options=_background_options(),
+        command_options=_command_options(),
     )
 
 
@@ -129,3 +138,9 @@ def _background_options() -> BackgroundOptions | None:
     if not settings.get_background():
         return None
     return BackgroundOptions(even_with_video=settings.get_background_always())
+
+
+def _command_options() -> CommandOptions | None:
+    if not settings.get_command():
+        return None
+    return CommandOptions(command=settings.get_command())
--- a/src/usdb_syncer/download_options.py
+++ b/src/usdb_syncer/download_options.py
@@ -66,7 +66,7 @@ class BackgroundOptions:
 class CommandOptions:
     """Settings regarding the command to be executed."""
 
-    command: settings.Command
+    command: str
 
 
 @dataclass(frozen=True)
@@ -143,4 +143,4 @@ def _background_options() -> BackgroundOptions | None:
 def _command_options() -> CommandOptions | None:
     if not settings.get_command():
         return None
-    return CommandOptions(command=settings.get_command())
+    return CommandOptions(command=settings.get_command_text())
--- a/src/usdb_syncer/gui/forms/SettingsDialog.ui
+++ b/src/usdb_syncer/gui/forms/SettingsDialog.ui
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>532</width>
-    <height>485</height>
+    <height>524</height>
    </rect>
   </property>
   <property name="windowTitle">
@@ -24,24 +24,11 @@
      </property>
     </widget>
    </item>
-   <item row="2" column="0">
-    <spacer name="verticalSpacer_2">
-     <property name="orientation">
-      <enum>Qt::Orientation::Vertical</enum>
-     </property>
-     <property name="sizeType">
-      <enum>QSizePolicy::Policy::Fixed</enum>
-     </property>
-     <property name="sizeHint" stdset="0">
-      <size>
-       <width>20</width>
-       <height>40</height>
-      </size>
-     </property>
-    </spacer>
-   </item>
-   <item row="0" column="0" rowspan="2">
+   <item row="0" column="0">
     <widget class="QTabWidget" name="tabWidget">
+     <property name="mouseTracking">
+      <bool>true</bool>
+     </property>
      <property name="currentIndex">
       <number>0</number>
      </property>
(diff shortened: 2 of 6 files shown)


randompersona1:
Oh, I couldn't figure out a way to create a test for this. We could split creating the args into a seperate function and test just that, but it honestly seemed pretty overkill for 4 lines of fairly simple code. If this got any more complicated in the future, I would just create tests then.

RumovZ:
I see multiple issues with running arbitrary shell commands:
- It's platform-specific, making script sharing and support more difficult.
- We're spawning a separate process that's competing for the same resources (files, DB access etc.).
- Code from the syncer cannot be reused, nor can popular scripts be easily integrated into the syncer.

I'd rather have a Python API giving us more control over the executed code, and the user all the logic already existing for the syncer at their disposal.
I imagine a hook system where user modules can subscibe to and execute arbitrary code at certain points. But that's just what I know from other projects and there may be better options.

randompersona1:
Thats fair. I do think this implementation has some advantages though: 
- It doesn't require python knowledge: you can use bash, powershell, or anything else
- The barrier of entry is extremely low compared to a hook-based system, requiring basically zero instructions or knowledge of how the syncer works internally. 
- Most importantly, it fits my use case: running a python3.11 script. If we used hooks, I would literally just write this code again to run an external command after a song has downloaded. That's not really a problem, but is kind of redundant to me personally.

To 1: I actually didn't really think of that. However, I would guess that users would write their scripts in a separate file and then run that. It's then at the users discretion if they want to use cross-platform languages or powershell/bash

And while hooks are certainly more elegant, I don't think they actually give us any more control. They still require resources, etc. Only thing I can think of is allowing direct log access. That would be technically possible with this implementation using pipes or files to transmit information back to the syncer, but I haven't been able to make it work reliably.

randompersona1:
This isn't production ready at all, but for reference, this is what I am using: https://github.com/randompersona1/songconvert

RumovZ:
> It doesn't require python knowledge: you can use bash, powershell, or anything else

It seems more reasonable to require Python than any other language not native to this project.

> I don't think they actually give us any more control. They still require resources, etc. 

No, they can reuse existing resources like the DB connection, which a separate process would not be able to acquire.

In the end, the most important thing is that the solution is universal. I understand that this would be most suitable for your use case, but the next user might need access to internal data or run code after some other event, and we can't add a custom solution for every use case. The solution must be easily extensible and maintainable.

randompersona1:
I realize it's unfair to impose my needs on everyone. That wasn't my intention. 

My main inspiration for this came from other programs that use similar features like handbrake.
image

The most common hook/ event would certainly be essentially this one - changing the downloaded files in some way after everything is downloaded, like reencoding the video with h.265 to save storage. My reasoning for doing it this way is simplicity for the user - they can just enter an ffmpeg command into the text box. Simultaneously, it would allow advanced users to still do basically anything because all important information is there in the completed folder. Slower than through the database? Absolutely. But also a lot more intuitive because the user can just look at the files and see all the information they've got. Text file parsing on a remotely modern system is instant anyways, so I doubt this would hamper performance in any significant way. And since the command is run inside the _run_inner(), it's multithreaded as well. 

I actually take zero issue with a purely python-based hook/event approach. That would still work great for me. I could start implementing such a solution if that is preferred. However, I am not sure the extra effort in coding, validating all kinds of inputs/variables after an event and documenting the events and their edge cases is worth it for the less coding-minded users that might not be able to take advantage of such an implementation anyways.

randompersona1:
Hey @RumovZ , what is the status? Do you want me to close or are you testing? Is somebody working on a hook-based system?

RumovZ:
@bohning, what do you think about providing an API? I always thought it would make sense eventually, but I wonder if the project is stable enough, yet.

@randompersona1, if you want to tackle the hook approach, maybe you could discuss here first what you have in mind.

randompersona1:
Okay, so I've thought about this a bit. Running something post download is fairly obvious. There could also be the capability for adding custom log handlers, replacing usdb with something else (other website, local backup), overriding default download functions. Theoretically, it could also be possible to populate existing or new table columns with a different data source. 

To be honest though, I doubt there's a lot more, and those ideas are already very niche. The development cost also seems quite high to me. I might think differently if there was an outspoken need, but it seems like i was the very first one to bring up this kind of extendability at all, even in the limited form that is this PR. 

Personally, I don't have the time to develop this for the next couple of weeks, but i could still take a shot at it after that. Generally, I am thinking there would be a plugins folder somewhere that the syncer imports at runtime. We'd also need a kind of SyncerPluginInterface that plugins implement for the syncer to call at the correct times. Since the syncer is built, I don't know how/if overriding methods would work, but I suppose checking would be possible. The most difficult part is type checking during development and I genuinely don't know how to solve it. If a plugin function takes a Context, how does anybody know what that is or what to with it without checking the source? Unless you're thinking about building the syncer with plugins, at which point the entire idea seems pretty redundant

bohning:
I am a little undecided here. The world is colorful and may be the needs for customizing post-processing karaoke songs, but on the other hand the goal of the Syncer was to make stuff as easy as possible for the normal user, offering as many settings as necessary and at the same time as few as possible in order to harmonize song collections in order to rule out differences (e.g. converting to mp3 may introduce a small delay, requiring a different #GAP).

So on the one hand, I think it would be a nice addition, on the other hand I am unsure if it is worth all the extra effort and if we should instead focus on integrating sensible functionality directly.

RumovZ:
I've added an API: https://github.com/bohning/usdb_syncer/tree/hooks
Does this work for you, @randompersona1?

> So on the one hand, I think it would be a nice addition, on the other hand I am unsure if it is worth all the extra effort and if we should instead focus on integrating sensible functionality directly.

On upside to an add-on system is that we explore different features without implementing them in the core codebase directly. If someone writes an add-on that proves useful and gets popular, we an easily integrate it later on.

randompersona1:
Nice work. Seems to work perfectly. I did notice namespace issues: The module name of the addon is limited and problems fail silently. That's obviously implementation specific though, so i'll close this.

PR #278, opened 2024-08-24
Add usdb as cover fallback

### LINKED ISSUE #212: Use USDB cover as fallback if metatag resource is not available
The USDB cover is already used as a fallback if there is *no* cover resource in the metatags. But it is not yet used as a fallback when the resource is not available.

randompersona1:
Fixes #212 

download_image() now accepts a list of urls and uses the first that works. We just add the usdb cover link to that list.

### DIFF
--- a/src/usdb_syncer/resource_dl.py
+++ b/src/usdb_syncer/resource_dl.py
@@ -145,33 +145,39 @@ def _download_resource(options: YtdlOptions, resource: str, logger: Log) -> str 
             return None
 
 
-def download_image(url: str, logger: Log) -> bytes | None:
-    try:
-        reply = requests.get(
-            url, allow_redirects=True, headers=IMAGE_DOWNLOAD_HEADERS, timeout=60
-        )
-    except requests.exceptions.SSLError:
-        logger.error(
-            f"Failed to retrieve {url}. The SSL certificate could not be verified."
-        )
-        return None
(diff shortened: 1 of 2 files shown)


Baklap4:
Can we create a static image telling the user to submit one if they encounter this song?
This might incentivize someone to upload a nice album cover

Then again, maybe that's not for USDB_Syncer to do, but rather the karaoke-programs if they happen to not see a cover attached

src/usdb_syncer/resource_dl.py
```diff
@@ -184,7 +190,14 @@ def download_and_process_image(
     max_width: CoverMaxSize | None,
 ) -> Path | None:
     logger = get_logger(__file__, details.song_id)
-    if not (img_bytes := download_image(url, logger)):
+
+    urls = (
+        [url, details.cover_url]
+        if kind == ImageKind.COVER and details.cover_url
+        else [url]
+    )
```
RumovZ, line 198: I would prefer to handle this in `song_loader`. It's already odd that we pass in `SongDetails` here at all. Would be cleaner to pass in the logger.

RumovZ: In fact, there is already `_Context.cover_url()` for handling the case where there's no cover URL in the meta tags. So this code should be modified instead of adding a loop later on.

PR #280, opened 2024-08-26
PathTemplate - retrieve first value of Genre, Language, Creator

Baklap4:
### Description
Retrieves either the first value or the whole string when no comma seperator has been found.

Using `partition()` instead of index-accessor + split since we don't need the rest of the string and this way we cannot go out of bounds if there is not match found.

### Usecase

I'm having a path-template setup like:
```
:language: / :artist: / :artist: - :title: / :artist: - :title:
```
For this [song](https://usdb.animux.de/?link=detail&id=29243) it results into
```
Japanese (romanized), English\Jujutsu Kaisen (ALI feat. AKLO)\Jujutsu Kaisen (ALI feat. AKLO) - Lost In Paradise (TV)\Jujutsu Kaisen (ALI feat.txt
```
With this fix it turns out to be:
```
Japanese (romanized)\Jujutsu Kaisen (ALI feat. AKLO)\Jujutsu Kaisen (ALI feat. AKLO) - Lost In Paradise (TV)\Jujutsu Kaisen (ALI feat.txt
```

It'd be nice to only keep the first entry of the comma-seperated list as path modifier.

It applies to the 3 mentioned fields: Genre, Language and Creator

### DIFF
--- a/src/usdb_syncer/path_template.py
+++ b/src/usdb_syncer/path_template.py
@@ -163,13 +163,16 @@ class PathTemplatePlaceholder(PathTemplateComponentToken, enum.Enum):
             case PathTemplatePlaceholder.TITLE:
                 return song.title
             case PathTemplatePlaceholder.GENRE:
-                return song.genre
+                firstGenre, _, _ = song.genre.partition(',')
+                return firstGenre.strip()
             case PathTemplatePlaceholder.YEAR:
                 return str(song.year)
             case PathTemplatePlaceholder.LANGUAGE:
-                return song.language
+                firstLanguage, _, _ = song.language.partition(',')
+                return firstLanguage.strip()
             case PathTemplatePlaceholder.CREATOR:
-                return song.creator
+                firstCreator, _, _ = song.creator.partition(',')
+                return firstCreator.strip()
             case PathTemplatePlaceholder.EDITION:
                 return song.edition
             case PathTemplatePlaceholder.RATING:
--- a/src/usdb_syncer/path_template.py
+++ b/src/usdb_syncer/path_template.py
@@ -163,16 +163,16 @@ class PathTemplatePlaceholder(PathTemplateComponentToken, enum.Enum):
             case PathTemplatePlaceholder.TITLE:
                 return song.title
             case PathTemplatePlaceholder.GENRE:
-                firstGenre, _, _ = song.genre.partition(",")
-                return firstGenre.strip()
+                first_genre, _, _ = song.genre.partition(",")
+                return first_genre.strip()
             case PathTemplatePlaceholder.YEAR:
                 return str(song.year)
             case PathTemplatePlaceholder.LANGUAGE:
-                firstLanguage, _, _ = song.language.partition(",")
-                return firstLanguage.strip()
+                first_language, _, _ = song.language.partition(",")
+                return first_language.strip()
             case PathTemplatePlaceholder.CREATOR:
-                firstCreator, _, _ = song.creator.partition(",")
-                return firstCreator.strip()
+                first_creator, _, _ = song.creator.partition(",")
+                return first_creator.strip()
             case PathTemplatePlaceholder.EDITION:
                 return song.edition
             case PathTemplatePlaceholder.RATING:
--- a/src/usdb_syncer/path_template.py
+++ b/src/usdb_syncer/path_template.py
@@ -163,15 +163,15 @@ class PathTemplatePlaceholder(PathTemplateComponentToken, enum.Enum):
             case PathTemplatePlaceholder.TITLE:
                 return song.title
             case PathTemplatePlaceholder.GENRE:
-                firstGenre, _, _ = song.genre.partition(',')
+                firstGenre, _, _ = song.genre.partition(",")
                 return firstGenre.strip()
             case PathTemplatePlaceholder.YEAR:
                 return str(song.year)
             case PathTemplatePlaceholder.LANGUAGE:
-                firstLanguage, _, _ = song.language.partition(',')
+                firstLanguage, _, _ = song.language.partition(",")
                 return firstLanguage.strip()
             case PathTemplatePlaceholder.CREATOR:
-                firstCreator, _, _ = song.creator.partition(',')
+                firstCreator, _, _ = song.creator.partition(",")
                 return firstCreator.strip()
             case PathTemplatePlaceholder.EDITION:
                 return song.edition
--- a/src/usdb_syncer/path_template.py
+++ b/src/usdb_syncer/path_template.py
@@ -163,16 +163,13 @@ class PathTemplatePlaceholder(PathTemplateComponentToken, enum.Enum):
             case PathTemplatePlaceholder.TITLE:
                 return song.title
             case PathTemplatePlaceholder.GENRE:
-                first_genre, _, _ = song.genre.partition(",")
-                return first_genre.strip()
+                return next(iter(song.genres()), "")
             case PathTemplatePlaceholder.YEAR:
                 return str(song.year)
             case PathTemplatePlaceholder.LANGUAGE:
-                first_language, _, _ = song.language.partition(",")
-                return first_language.strip()
+                return next(iter(song.languages()), "")
             case PathTemplatePlaceholder.CREATOR:
-                first_creator, _, _ = song.creator.partition(",")
-                return first_creator.strip()
+                return next(iter(song.creators()), "")
             case PathTemplatePlaceholder.EDITION:
                 return song.edition
             case PathTemplatePlaceholder.RATING:
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ venv
 # IDEs
 .qt_for_python
 .vscode
+.vs
 
 # tooling
 .tox


Baklap4:
@bohning do you want me to squash/fixup the commits since the last two are to satisfy Tox or would a history like this be alright?

src/usdb_syncer/path_template.py
```diff
@@ -163,13 +163,16 @@ def evaluate(self, song: UsdbSong) -> str:
             case PathTemplatePlaceholder.TITLE:
                 return song.title
             case PathTemplatePlaceholder.GENRE:
-                return song.genre
+                firstGenre, _, _ = song.genre.partition(',')
```
bohning: Why not use `return song.genre.split(',')[0]?

Baklap4: Couple of reasons:
1. No out of bounds check needed cause index 0 could be not set in the case of split with an empty string leading to IndexError
2. Predictable output. Partition always returns a tuple of 3 values: <Item before the comma> | <the comma> | <rest of string>. As mentioned in the PR we don't need the other 2 values
3. Performance wise; split enumerates the whole array whereas partition stops at the first occurence
4. Readability

randompersona1:
I wouldn't actually do this for genres. Separating between "Pop, Rock" and "Pop, Funk" seems okay. Much worse for genres is how there is no defined list of genres, thus leading to many different separators. I see "," "/" "-" " " just for pop songs. What about just removing those and potentially any other separators, like usdb seems to for the sorting?

Baklap4:
> I wouldn't actually do this for genres. Separating between "Pop, Rock" and "Pop, Funk" seems okay. Much worse for genres is how there is no defined list of genres, thus leading to many different separators. I see "," "/" "-" " " just for pop songs. What about just removing those and potentially any other separators, like usdb seems to for the sorting?

[The format project defines that values are comma-seperated](https://github.com/UltraStar-Deluxe/format/blob/main/spec.md#320-the-genre-header)
I added it for Genre, Languages and Creator since these 3 were mentioned within the discord.

randompersona1:
The format defines that \#GENRE is a header and that it can be multi-valued. On usdb, there are loads of songs with the separators i metioned, e.g "Pop-Rock", "Pop Rock", "Pop, Rock", etc. 

The real problem is that there is no list of genres to choose from to make a comma-separated list ([if needed, musicbrainz has an extensive list](https://musicbrainz.org/genres)), but a stopgap solution would be to remove the separators, as usdb seems to do.

Baklap4:
> The format defines that #GENRE is a header and that it can be multi-valued. On usdb, there are loads of songs with the separators i metioned, e.g "Pop-Rock", "Pop Rock", "Pop, Rock", etc.
> 
> The real problem is that there is no list of genres to choose from to make a comma-separated list ([if needed, musicbrainz has an extensive list](https://musicbrainz.org/genres)), but a stopgap solution would be to remove the separators, as usdb seems to do.

Again, the format specifies that multivalued items are comma seperated. See https://github.com/UltraStar-Deluxe/format/blob/main/spec.md#2-the-file-header and  https://github.com/UltraStar-Deluxe/format/blob/main/spec.md#31-single-valued-and-multi-valued-headers

The fact that USDB lists songs with other formats is not the debate here. In the end we're trying to get usdb up to standard as much as possible, and changes are being made to help us with that. Yet they come in rather slow as there's just a single developer working on that.

The list of genres can be fetched from anywhere by all means it could be gibberish, the only fact is, it should be comma seperated which is what this PR is about and will support

randompersona1:
This is a stupid argument. I am pointing out two things the PR doesn't address, but might not need to if this is the behaviour you intended.
1. a lot of songs on usdb right now use wierd separators for the genre, like "-".
2. Putting "Pop, Rock" into the "Pop" folder without notice might be confusing or unwanted behaviour, particularly when combined with 1, where "Pop-Funk" gets it's own folder while "Pop, Rock" doesn't

I have no intention of dying on this hill, so do with this whatever you want.

Baklap4:
I merely added genre simply because it was mentioned in the discord. You could argue if we'd want to put "Pop, Rock" into a folder called "Pop". Agreed on that front. Up to @bohning i guess

Do you have a couple of songs using the `-` seperator where it is actually meant to be 2 different genres and not a merger.? because most genres i'm seeing are correctly using the `-` for example: `J-Pop` or `K-Pop` which uses `-` but is a genre of its own

The ones with multi-valued genres are correctly using the comma to distinguish between them as far as the 20 pages i just scrolled through

randompersona1:
https://usdb.animux.de/?link=detail&id=7626
https://usdb.animux.de/?link=detail&id=22165
There are loads more. Just sort by genre and look around.

Unless there there is a musical difference between the genre "Pop Rock" and listing "Pop" and "Rock" as genres. tbh, i'm pretty sure the people that uploaded don't know either, making it impossible to guess how it was intended.

bohning:
IMHO the entire genre topic is a hot one. As long as there is no defined way, it will be a mess. Is "Pop-Rock" a genre on its own, and is it the same as "Rock-Pop", or is "Rock-Pop" more Rock than Pop and "Pop-Rock" is the other way round? And is "Pop, Rock" yet different from that (and what about "Rock, Pop" for the sake of completeness for this example alone). I tried to harmonize this a bit (merging e.g. JPop, Jpop, J-pop, j pop, J Pop, JPOP, ... and J-Pop to the latter, same for C-Pop and K-Pop which seem to be established genres on their own), but then we also have language specific genres like (German) Schlager, Deutsch-Rock (should that be translated to German Rock?) etc. There are high level genres (Pop, Rock, Metal, ...) and more specific genres (Punk Rock/Punk-Rock, Psychedelic Rock, Glam Rock)... So I'd say the best solution is to find a comprehensive catalog to look it up (MusicBrainz seems to be the obvious choice), but AFAIK it provides a list of genres per artist (not per song). They can be added/upvoted by users, so maybe taking the top three genres or the genres with more than x votes could be sensible. Otherwise, I think the #GENRE tag will always be a mess. The developer of UltraSinger pointed me towards https://github.com/jordipons/musicnn, which could be an alternative to MusicBrainz.

Baklap4:
@bohning in case of this pr (and thus skipping all the specifics for genre in this case), should we keep it as is? Or rather delete my change in regard to Genre?

My main goal was to have the main language of a song as a folder instead of a comma separated list, that can be achieved by this.
The other multivalued fields i'm not using but i can guess other people might use it. Aside they were mentioned in discord so out of courtesy added them in, but that can be removed to fix my specific usecase

bohning:
For the sake of the above argument ("genre is a hot topic/a mess"), here's 81 different variants of genres that contain the word "rock" that are currently present in usdb:

Comedy Rock
Acoustic Rock
Alternative Rock
Art Rock
Blues Rock
Boogie Rock
Celtic Punk/Rock
Christian Hard Rock
Christian Rock
Classic Pop-Rock
Classic Rock
Comedy Rock
Dance Rock
Dark Rock
Deutsch Rock Pop
Deutsch-Rock
Dino Rock
Doo Wop Rock & Roll
Electro Rock
Electronic Rock
Emo-Rock
Folk Rock
Folk-Rock
Funk Rock
Garage Rock
General Alternative Rock
General Mainstream Rock
Glam Rock
Gothic Rock
Hard Rock
Hard Rock & Metal
Indie Rock
Indie-Rock
Industrial Rock
J-Rock
Jazz-Rock
Latin Rock
Mittelalter Rock
Mittelalter-Rock
Modern Hard Rock
Mundart Rock
Ost-Rock
Piano Rock
Pop Rock
Pop-Rock
Pop-Rock Oldies
Post Rock
Progressive Rock
Psychedelic Rock
Pub Rock
Punk Rock
Rap Rock
Rap-Rock
RnB Rock
Rock
Rock & Metal
Rock & Roll
Rock Alternative
Rock Argentino
Rock Ballad
Rock Catalán
Rock Celtic
Rock Indie
Rock Industrial
Rock Latino
Rock Psicodelico
Rock Schlager
Rock ’n Roll
Rock ’n’ Roll
Rock&Roll
Rock-Pop
Rock’n’Roll
Shagadelic Rock
Ska Rock
Soft Rock
Southern Rock
Stoner Rock
Surf Rock
Symphonic Rock
Symphonic-Rock-Cover
Synth Rock

src/usdb_syncer/path_template.py
Baklap4: If this is alright i'll resolve the convo :)

src/usdb_syncer/path_template.py
```diff
@@ -163,13 +163,16 @@ def evaluate(self, song: UsdbSong) -> str:
             case PathTemplatePlaceholder.TITLE:
                 return song.title
             case PathTemplatePlaceholder.GENRE:
-                return song.genre
+                first_genre, _, _ = song.genre.partition(",")
```
RumovZ, line 166: I would prefer to use the existing accessors to avoid duplicating the splitting logic, e.g. `next(song.genres()) or ""`.

RumovZ:
I can see that genres are a difficult topic, but until there's further specification, @Baklap4's current implementation makes the most sense to me. Putting a "Rock, Pop" song into a "Rock" folder might be less than ideal, but more acceptable than folders like "Rock, Electronic, Pop, Electro, Funk, Synth-Pop, Pop Rock, Funk Soul" (real example).

src/usdb_syncer/path_template.py
Baklap4: Changes have been applied 👍  also made it up-to-date with latest main

Baklap4:
@RumovZ I added a line to `gitignore` ignoring the `.vs`  folder. Also rebased it, and removed the actual empty file/folder.

RumovZ (approved):

Issue #167, opened 2023-09-19
Comment and/or rate via Syncer


bohning:
The Syncer could provide the functionality to add a comment to a song on usdb or give a 1-5 star rating.

Baklap4:
Partially implemented by #282 
Haven't taken Rating in account

Baklap4:
Rating of the comment (neutral/positiv/negativ) have been added to the pr

PR #282, opened 2024-09-04
Post comments onto songs

Baklap4:
### Posting a comment through USDB-Syncer

In this PR i built it so that you can post a comment on one or multiple songs.
Right-click your song(s) and hit the `Post Comment` submenu item.
You've got to fill in the comment, and hit the save button.
For each song the comment will be posted under your current logged in user.

### Usecases

* Adding Video content
* Adding Tags
* Probably other usecases

### Implements

* Partially implements #167 

### My usecase
* Adding anime names as `#TAGS` to multiple songs so i can search them later.

### Personal note

I've allowed edits and access, so feel free to update the code accordingly it's a first crude attempt with my minor python and QT knowledge :P I basicly copied and pasted a lot of stuff from another dialog (UsdbLoginDialog)

### Screenshots

image

image

image


src/usdb_syncer/gui/forms/CommentDialog.py
```diff
@@ -0,0 +1,75 @@
+# -*- coding: utf-8 -*-
```
RumovZ, line 1: This file is generated and should not be checked in.

src/usdb_syncer/gui/comment_dialog.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def accept(self) -> None:
+        for song in self._selected_songs():
+            payload = {"text": self.line_edit_username.text(), "stars": "neutral"}
+
+            get_usdb_page(
+                "index.php",
+                RequestMethod.POST,
+                headers={"Content-Type": "application/x-www-form-urlencoded"},
+                params={"link": "detail", "id": str(song.song_id), "comment": str(1)},
```
RumovZ, line 31: This might need to be `str(int(song.song_id))` if leading 0s cause problems.

src/usdb_syncer/gui/comment_dialog.py
```diff
@@ -0,0 +1,34 @@
...
+from usdb_syncer.gui.forms.CommentDialog import Ui_Dialog
+from usdb_syncer.usdb_scraper import RequestMethod, get_usdb_page
+from usdb_syncer.usdb_song import UsdbSong
+
+
+class CommentDialog(Ui_Dialog, QDialog):
+    """Dialog to post a comment on one or multiple songs."""
+
+    def __init__(
+        self, parent: QWidget, selected_songs: Callable[[], Iterator[UsdbSong]]
```
RumovZ, line 16: Would be better to just accept a list or iterator of song ids.

src/usdb_syncer/gui/forms/CommentDialog.ui
```diff
@@ -0,0 +1,99 @@
...
+				<layout class="QFormLayout" name="formLayout">
+					<item row="0" column="0">
+						<widget class="QLabel" name="label">
+							<property name="text">
+								<string>Comment:</string>
+							</property>
+						</widget>
+					</item>
+					<item row="0" column="1">
+						<widget class="QLineEdit" name="line_edit_username">
```
RumovZ, line 27: Might be nicer to have a text area instead of a single line.

RumovZ (commented):
Cool feature! I'm just a bit worried that it's also practically a spamming tool. You could even post comments to thousands of songs by mistake if you're not careful.

Baklap4:
> Cool feature! I'm just a bit worried that it's also practically a spamming tool. You could even post comments to thousands of songs by mistake if you're not careful.

Agreed. It is limited (not by code but the api) that one needs to be logged in.
But aside from that it might result into spamming a lot of comments.

I did test it out adding a couple of tags to a multitude of songs. But i mostly end up just selecting a single song as the tags won't apply to every selected song..

Shall we bring this down to just a single selected item?

src/usdb_syncer/gui/forms/CommentDialog.py
Baklap4: How is this generated? I might not have the tooling for this installed

src/usdb_syncer/gui/comment_dialog.py
Baklap4: I'll test this out, thanks for the good catch

src/usdb_syncer/gui/comment_dialog.py
Baklap4: Yeah my first implementation was to accept a list, but then i made it callable so we don't have the method execution in mw.py. I can revert that, thanks

src/usdb_syncer/gui/forms/CommentDialog.ui
Baklap4: Gotta find the widget for this, and it seems i forgot to alter the name aswell

Would the class be a `QTextEdit`? Or is it another class?

src/usdb_syncer/gui/forms/CommentDialog.py
RumovZ: It already got generated. The issue is that it's checked in. You have to `git rm` it.

src/usdb_syncer/gui/forms/CommentDialog.ui
RumovZ: There's also QPlainTextEdit, I'm not sure what's appropriate here.

RumovZ:
> Shall we bring this down to just a single selected item?

Yes, I think that would mostly eliminate the misuse potential. Or what do you think, @bohning?

bohning:
> > Shall we bring this down to just a single selected item?
> 
> Yes, I think that would mostly eliminate the misuse potential. Or what do you think, @bohning?

Yes, I would agree. In general, it should be used to leave a real comment for a particular song, so limiting it to a single song seems most sensible to me.

Baklap4:
I'll get to this next week, thanks for the feedback :)

For now the comment is neutral, do we want it so the user can specify it?

bohning:
> For now the comment is neutral, do we want it so the user can specify it?

I'd say yes, please.

RumovZ:
> Also do we want a specific image for comments or is the Price-Tag png good enough?

A unique icon would be better. You can choose one from here (make sure to pick the shadowless variant): https://p.yusukekamiyamane.com/

src/usdb_syncer/gui/forms/CommentDialog.py
Baklap4: Removed this file

src/usdb_syncer/gui/comment_dialog.py
Baklap4: This sends the request with leading zero's. It's not a problem. So now it becomes a matter of what we want? I do see that USDB itself doesn't use the leading zero's in their routing.

src/usdb_syncer/gui/comment_dialog.py
Baklap4: Fixed this and it's also just a single song now.

src/usdb_syncer/gui/forms/CommentDialog.ui
Baklap4: Changed it to QPlainTextEdit since we're not sending formated text as far as i know (html/markdown etc)

src/usdb_syncer/gui/comment_dialog.py
Baklap4: Changed it in the end, since usdb can't load the cover when sending leading zeros

Baklap4:
Aight i've updated the PR with the request changes. In summary:

* Added unique comment icon
* Changed textfield to textedit
* Added rating dropdown
* changed from multi song selection to just one song

src/usdb_syncer/gui/forms/CommentDialog.ui
bohning: USDB supports HTML in comments, so theoretically it could be formatted, not sure if we want that.

Baklap4: Up to you guys if we want it supported (it's easy enough to add in)

bohning: I'd say leave it out for now, it can be extended later if it is really needed (and users can manually use HTML code as comments if it is really needed).

src/usdb_syncer/gui/comment_dialog.py
```diff
@@ -0,0 +1,39 @@
...
+    """Dialog to post a comment on one or multiple songs."""
+
+    def __init__(self, parent: QWidget, selected_song: UsdbSong) -> None:
+        super().__init__(parent=parent)
+        self._parent = parent
+        self.setupUi(self)
+        self._selected_song = selected_song
+        self.combobox_rating.addItem("neutral")
+        self.combobox_rating.addItem("positiv")
+        self.combobox_rating.addItem("negativ")
```
RumovZ, line 20: Could we make that English labels, or icons?

RumovZ:
Also, might be worth mentioning in the changelog.

Baklap4:
> Also, might be worth mentioning in the changelog.

How would i go about the changelog? I don't see it mentioning any specific version but rather a copy of the whole 'changelog' block. For now i can just copy that block and add in the details... 

... But it might be good to look into a structured changelog which points changes to specific versions -> https://keepachangelog.com/ If that's something you'd like to see i'll do that in another PR

bohning:
> How would i go about the changelog?

Just add the comment functionality to 0.8.0 here: https://github.com/bohning/usdb_syncer/blob/main/CHANGELOG.md

RumovZ:
The versions are in the comments. You have to look at the raw text: https://raw.githubusercontent.com/bohning/usdb_syncer/main/CHANGELOG.md

src/usdb_syncer/gui/comment_dialog.py
RumovZ: Just add another argument to `addItem()`. The second one will be retrievable with `currentData()`.

Baklap4: Added the labels (as text)

Baklap4:
Ahh that's why it didn't show up in the preview of the markdown
I've added:
* The labels Positive, Negative, Neutral. I'm still sending the probably German values as this is what USDB expects
* Updated the changelog containing a brief description of this feature

Thanks for the great feedback!

RumovZ (approved):

PR #296, opened 2024-10-11
Bugfixes

bohning:
Some small bugfixes/improvements:
- #TAGS is now only written to the text file if it is not empty (we should probably migrate to a `tags=` metatag now with the basically unlimited line length on USDB)
- Add some explanatory text to the legacy support text encoding options UTF-8 BOM and CP1252
- Disable and hide the recently introduced option to add artwork to mp4 as this produces video files that are currently not playable by USDX

### DIFF
--- a/src/usdb_syncer/gui/forms/SettingsDialog.ui
+++ b/src/usdb_syncer/gui/forms/SettingsDialog.ui
@@ -416,6 +416,9 @@
             </item>
             <item row="3" column="0">
              <widget class="QLabel" name="label_video_embed_artwork">
+              <property name="enabled">
+               <bool>false</bool>
+              </property>
               <property name="text">
                <string>Embed artwork:</string>
               </property>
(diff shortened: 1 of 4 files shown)


RumovZ (approved):

RumovZ:
> we should probably migrate to a tags= metatag now with the basically unlimited line length on USDB

I think another reason it's currently not a metatag are the comma separators. You'd have to remember encoding them every time.

bohning:
Yes, but we could specify that multiple tags should be e.g. hyphen-separated (like medley, crop), e.g. `tags=explicit-duet`.

RumovZ:
The spec only mentions commas, as far as I'm aware, so every other symbol may be part of a tag. Specifically, if we use hyphens, a tag like `Rock-Pop` would need to be encoded.

bohning:
I meant using hyphens for metatags only, converting it back to commas when writing the actual #TAGS. But then a hyphen is not ideal as it can occur in tags (

RumovZ:
> I meant using hyphens for metatags only

I know.

> Maybe "|" as separator?

As I said, except for `,` everything may be part of tag. Of course, we can pick some obscure separator and hope no one ever adds a tag containing it in practice. But that's guesswork and not a valid format.
However, I just realized our encoding is broken anyway, because we only replace with `,` with `%2C` and there's nothing from preventing `%2C` itself to appear in a metatag value. Well, so much for a valid format. 🙂

PR #301, opened 2024-10-20
Add possibility to open karaoke apps for the current song.

bohning:
Unsolved issue:
- pressing Cancel on the FileDialog to select the executable closes the entire settings screen and it does not clear the lineEdit for the executable path (Mac only, it seems).


src/usdb_syncer/gui/mw.py
```diff
@@ -281,6 +311,21 @@ def _open_current_song_folder(self) -> None:
...
+                else:
+                    with db.transaction():
+                        song.remove_sync_meta()
+                    events.SongChanged(song.song_id)
+                    _logger.info("Song does not exist locally anymore.")
+            else:
+                _logger.info("Song does not exist locally.")
+        else:
+            _logger.info("No current song.")
+
```
RumovZ, line 328: This function is an almost identical duplicate. The common parts should be refactored into a separate function.

RumovZ, line 328: This function is an almost identical duplicate. The common parts should be refactored into a separate function.

src/usdb_syncer/gui/resources/resources.qrc
```diff
@@ -1,5 +1,13 @@
 <RCC>
   <qresource prefix="icons">
+    <file>music--arrow.png</file>
+    <file>karedi.png</file>
+    <file>vocaluxe.png</file>
+    <file>performous.png</file>
+    <file>ultrastar-manager.png</file>
+    <file>yass-reloaded.png</file>
+    <file>usdx.png</file>
+    <file>yass.png</file>
```
RumovZ, line 10: Not that important, but this was a sorted list until now.

src/usdb_syncer/gui/settings_dialog.py
```diff
@@ -42,6 +45,62 @@ def __init__(self, parent: QWidget, song: UsdbSong | None) -> None:
         self.label_video_embed_artwork.setVisible(False)
         self.checkBox_video_embed_artwork.setVisible(False)
         self.groupBox_reencode_video.setVisible(False)
+        if sys.platform != "win32":
+            self.groupBox_vocaluxe.setVisible(False)
+        self.pushButton_browse_karedi.clicked.connect(
+            lambda: self._set_location(settings.SupportedApps.KAREDI)
```
RumovZ, line 51: This saves the setting immediately. All other settings are only saved once the user presses Ok.

src/usdb_syncer/settings.py
```diff
@@ -487,6 +494,96 @@ def __str__(self) -> str:
...
+        elif executable.endswith(".app"):
+            cmd = [
+                os.path.join(executable, "Contents", "MacOS", self.executable_name()),
+                self.songpath_parameter(),
+                str(path),
+            ]
+        else:
+            cmd = [executable, self.songpath_parameter(), str(path)]
+        try:
+            subprocess.Popen(cmd)  # pylint: disable=consider-using-with
```
RumovZ, line 572: Might be nice to add a comment remembering us why we want to ignore this lint.

src/usdb_syncer/settings.py
```diff
@@ -754,3 +851,39 @@ def get_path_template() -> path_template.PathTemplate:
 
 def set_path_template(template: path_template.PathTemplate) -> None:
     set_setting(SettingKey.PATH_TEMPLATE, template)
+
+
+def get_app_path(app: SupportedApps) -> str:
```
RumovZ, line 856: Since this has to be a valid file path, we should use `Path`. (And maybe `None` to unset it?)

src/usdb_syncer/settings.py
```diff
@@ -487,6 +494,96 @@ def __str__(self) -> str:
...
+                f"File not found: {fnf_error}. "
+                "Please check the executable path in the settings."
+            )
+            traceback.print_exc()
+        except OSError as os_error:
+            _logger.error(f"OS error: {os_error}")
+            traceback.print_exc()
+        except subprocess.SubprocessError as sp_error:
+            _logger.error(f"Subprocess error: {sp_error}")
+            traceback.print_exc()
```
RumovZ, line 584: The explicit error handling is very nice.
However, `print` statements won't show up in the log, and we usually give the user a more friendly message than the raw exception, like your second line for `FileNotFoundError`.
For the other cases, you can search the code base for `format_exc` for examples.
Also, the path that caused the issue would be helpful.

src/usdb_syncer/gui/settings_dialog.py
```diff
@@ -181,4 +258,24 @@ def _save_settings(self) -> bool:
...
+        settings.set_app_path(
+            settings.SupportedApps.USDX, self.lineEdit_path_usdx.text()
+        )
+        settings.set_app_path(
+            settings.SupportedApps.VOCALUXE, self.lineEdit_path_vocaluxe.text()
+        )
+        settings.set_app_path(
+            settings.SupportedApps.YASS_RELOADED,
+            self.lineEdit_path_yass_reloaded.text(),
+        )
```
RumovZ, line 280: Might be nice to do some validation, i.e. check if the pass exists. Alternatively, make the text edits readonly so they can only be set to a valid path.

RumovZ (changes requested):

src/usdb_syncer/gui/settings_dialog.py
bohning: So just deleted `settings.set_app_path(app, path)`?

src/usdb_syncer/settings.py
bohning: Can I store a Path object in the settings, or do I convert to string before saving it to settings?

src/usdb_syncer/settings.py
bohning: So just a friendly message, and ignore the raw error? Or error with a friendly message and log raw exception as debug (in addition to traceback.format_exc)? Or does format_exc include the raw exception already?

RumovZ: `format_exc` gives you the traceback including the exception. In case the file wasn't found, we know exactly what went wrong and this won't be very interesting. But in case of more ambiguous or unexpected errors, logging any information we can will be helpful when debugging.

src/usdb_syncer/settings.py
bohning: Not sure if I did everything correctly here...

RumovZ:
> pressing Cancel on the FileDialog to select the executable closes the entire settings screen and it does not clear the lineEdit for the executable path.

Works for me. Maybe it's another Qt bug? Can you try with a different PySide version?

bohning:
> Works for me. Maybe it's another Qt bug? Can you try with a different PySide version?
Doesn't work for me, tried now with 6.7.3, 6.8.0.1, 6.7.0. and 6.6.0. So when you press Cancel, the line edit is emptied (if it was set before)?

RumovZ:
> So when you press Cancel, the line edit is emptied (if it was set before)?

Yes. Seems to be a bug on macOS.

bohning:
Seems like it. But I guess we can live with that.

RumovZ (approved):

PR #304, opened 2024-10-29
Only extract ID from actual YT URLs.

bohning:


src/usdb_syncer/constants.py
```diff
@@ -97,7 +97,7 @@ class Usdb:
     r"""\b
         (
             (?:https?://)?
-            (?:www\.)?
+            (?:www|web)\.?
```
RumovZ, line 100: This makes a leading www or web mandatory, which I assume was not the intent.

RumovZ (approved):

PR #306, opened 2024-10-30
Switch from browser_cookie3 to rookiepy.

bohning:


RumovZ:
I tried some browsers on Windows:

| | browser_cookie3 | rookiepy |
| --- | :---: | :---: |
| Chrome | ❌ |  ❌ |
| Firefox | ✔️ | ✔️ |
| Edge | ❌ | ✔️ (only admin)  |

So unfortunately, the only improvement is with Edge, provided the user is willing and able to run in admin mode. But at least that's something.

rookiepy has some annoying logging we should get under control before merging.

bohning:
I tried some browsers on Mac:

| | browser_cookie3 | rookiepy |
| --- | :---: | :---: |
| Chrome | ❌ |  ✔️ |
| Firefox | ✔️ | ✔️ |
| Edge | ❌ | ✔️ |
| Safari | ✔️ | ✔️ |

> So unfortunately, the only improvement is with Edge, provided the user is willing and able to run in admin mode. But at least that's something.
On Mac, it also works for Chrome. And rookiepy seems to be more actively developed.

> rookiepy has some annoying logging we should get under control before merging.
Yes.

RumovZ (approved):

PR #307, opened 2024-11-03
Optional fix for quotation marks

bohning:


src/usdb_syncer/song_txt/auxiliaries.py
```diff
@@ -42,14 +47,51 @@ def make_large_enough(self) -> int:
...
+    quotation_mark_indices: list[int] = []
+    for mark in QUOTATION_MARKS_TO_REPLACE:
+        quotation_mark_indices.extend(
+            match.start() for match in re.finditer(re.escape(mark), value)
+        )
+    quotation_mark_indices.sort()
+
+    marks_fixed = 0
+    for str_index in quotation_mark_indices:
+        marks_index = marks_total % 2
```
RumovZ, line 86: This doesn't take into account nested quotes. Not sure how common it would be in lyrics, but in theory you could have something like `he said 'hello "world"'`. So there is even a semantic difference between different kinds of quotation marks.

src/usdb_syncer/constants.py
```diff
@@ -164,3 +164,43 @@ class Usdb:
...
+        "Turkish",
+        "Vietnamese",
+    ): ("“", "”"),
+    ("Czech", "Estonian", "Icelandic", "Serbian", "Slovak", "Slovenian"): ("„", "“"),
+    ("Danish", "German"): ("»", "«"),
+    ("Japanese", "Japanese (romanized)"): ("「", "」"),
+    ("Welsh",): ("‘", "’"),
+}
+
+QUOTATION_MARKS_TO_REPLACE = {'"', "„", "”", "«", "»", "“", "「", "」"}
```
RumovZ, line 206: Is this the subset of all quotation marks above minus the ones that could also be an apostrophe?

src/usdb_syncer/constants.py
```diff
@@ -164,3 +164,43 @@ class Usdb:
...
+        "Irish",
+        "Korean",
+        "Korean (romanized)",
+        "Portuguese (Brazil)",
+        "Scots",
+        "Turkish",
+        "Vietnamese",
+    ): ("“", "”"),
+    ("Czech", "Estonian", "Icelandic", "Serbian", "Slovak", "Slovenian"): ("„", "“"),
+    ("Danish", "German"): ("»", "«"),
```
RumovZ, line 201: How did you come up with this list? E.g. why the "alternative" and not the "standard" quotes in German?

src/usdb_syncer/song_txt/auxiliaries.py
bohning: Agreed, I guess in theory this could occur, even though I cannot recall having seen that even once. But indeed the current implementation would screw things up for that case.

src/usdb_syncer/constants.py
bohning: »Im Amtlichen Regelwerk (z. B. Duden [..], ebenso im Wahrig [...]) werden die Regeln zur Verwendung behandelt, nicht jedoch, ob „deutsche“ Anführungszeichen oder Guillemets zu verwenden sind. Beide Möglichkeiten stehen gleichwertig nebeneinander. [...] Vorteile der Guillemets gegenüber den deutschen Anführungszeichen sind die Unverwechselbarkeit gegenüber Komma, Apostroph und Zollzeichen. Außerdem stören die deutschen Anführungszeichen die Zeilenbildung durch ihr aus der Zeilenmitte gerücktes Bild, die französischen Anführungen hingegen ähneln anderen Interpunktionszeichen nicht und unterstützen die Zeilenbildung. Nach Ansicht des Textportals Rotkel zeichnen sich die Guillemets außerdem durch eine elegantere Erscheinungsform aus.« (https://de.wikipedia.org/wiki/Anf%C3%BChrungszeichen#Deutschland_und_%C3%96sterreich).

In short: aesthetics ;-)

src/usdb_syncer/constants.py
bohning: Plus the typewriter (upright) quotation marks " (first entry).

src/usdb_syncer/song_txt/auxiliaries.py
RumovZ: If we want to take that risk, maybe we should at least have a comment that this case is intentionally unsupported.

src/usdb_syncer/song_txt/auxiliaries.py
```diff
@@ -42,14 +47,51 @@ def make_large_enough(self) -> int:
...
+            for langs, quote_marks in QUOTATION_MARKS.items()
+            if language in langs
+        ),
+        ("“", "”"),
+    )
+
+    quotation_mark_indices: list[int] = []
+    for mark in QUOTATION_MARKS_TO_REPLACE:
+        quotation_mark_indices.extend(
+            match.start() for match in re.finditer(re.escape(mark), value)
```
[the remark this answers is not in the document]
RumovZ: In fact, we don't seem to need regex at all.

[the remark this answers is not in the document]
bohning: Like this?

```
quotation_mark_indices = sorted(
        index
        for mark in QUOTATION_MARKS_TO_REPLACE
        for index in range(len(value))
        if value.startswith(mark, index)
)
```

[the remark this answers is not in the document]
RumovZ: Hm, I just realise, some replacements have more than one character. There are at least two issues with this:
1. If you replace one character with two, any following previously collected indices will be wrong.
2. `«` will be replaced with `« ` unconditionally, possibly leading to multiple spaces.

src/usdb_syncer/song_txt/auxiliaries.py
```diff
@@ -42,14 +47,51 @@ def make_large_enough(self) -> int:
...
+            value = (
+                value[:str_index]
+                + quotation_marks[marks_index]
+                + value[str_index + 1 :]
+            )
+            marks_fixed = marks_fixed + 1
+
+        marks_total = marks_total + 1
+
+    return value, marks_total, marks_fixed
```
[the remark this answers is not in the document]
RumovZ: Yes, that's one way.

RumovZ:
As the code is becoming more complicated, it's hard to tell if it does the right thing just from looking at it. Some unit tests would be helpful.

RumovZ (approved):

PR #313, opened 2024-11-11
Implement format version support.

bohning:


src/usdb_syncer/gui/forms/SettingsDialog.ui
```diff
@@ -103,72 +106,88 @@
...
+            <item row="1" column="1">
+             <widget class="QComboBox" name="comboBox_line_endings">
+              <property name="sizePolicy">
+               <sizepolicy hsizetype="Preferred" vsizetype="Fixed">
+                <horstretch>1</horstretch>
+                <verstretch>0</verstretch>
+               </sizepolicy>
+              </property>
+              <property name="toolTip">
+               <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;USDX and variants can handle both line ending types, but if you want to edit the song file manually, your editor of choice may required a certain line ending. UNIX users (MacOS, Linux) should generally use LF (line feed). On Windows, some text editors may require CRLF (carriage return, line feed). USDB currently requires CRLF line endings.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
```
RumovZ, line 118:
```suggestion
               <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;USDX and variants can handle both line ending types, but if you want to edit the song file manually, your editor of choice may require a certain line ending. UNIX users (MacOS, Linux) should generally use LF (line feed). On Windows, some text editors may require CRLF (carriage return, line feed). USDB currently requires CRLF line endings.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
```

src/usdb_syncer/settings.py
```diff
@@ -71,10 +71,11 @@ class SettingKey(Enum):
     TXT = "downloads/txt"
     ENCODING = "downloads/encoding"
     NEWLINE = "downloads/newline"
+    FORMAT_VERSION = "downloads/format_version"
     FIX_LINEBREAKS = "fixes/linebreaks"
-    FIX_FIRST_WORDS_CAPITALIZATION = "fixes/firstwordscapitalization"
+    FIX_FIRST_WORDS_CAPITALIZATION = "fixes/first_words_capitalization"
```
RumovZ, line 76: Every time you change the key, the user stored settings gets lost.

src/usdb_syncer/song_loader.py
```diff
@@ -518,14 +520,63 @@ def _maybe_write_txt(ctx: _Context) -> None:
...
+        case FormatVersion.V1_0_0:
+            ctx.txt.headers.mp3 = path.name
+        case FormatVersion.V1_1_0:
+            # write both #MP3 and #AUDIO to maximize compatibility
+            ctx.txt.headers.mp3 = path.name
+            ctx.txt.headers.audio = path.name
+        case FormatVersion.V1_2_0:
+            ctx.txt.headers.audio = path.name
+            if resource := ctx.txt.meta_tags.audio or ctx.txt.meta_tags.video:
+                ctx.txt.headers.audiourl = video_url_from_resource(resource)
```
RumovZ, line 553: Should assert exhaustive matching.

src/usdb_syncer/settings.py
bohning: I was aware of that, but since it was only recently introduced and I noticed that this key does not follow the other key spellings, I decided to change it. But I can change that back to the way it was if you prefer.

src/usdb_syncer/gui/forms/SettingsDialog.ui
```diff
@@ -103,72 +106,88 @@
...
               </property>
              </widget>
             </item>
             <item row="2" column="1">
-             <widget class="QComboBox" name="comboBox_fix_linebreaks">
-              <property name="toolTip">
-               <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;If this setting is enabled, the linebreak timings will be recalculated, either according to USDX or to YASS rules.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
-              </property>
-             </widget>
+             <widget class="QComboBox" name="comboBox_format_version"/>
```
RumovZ, line 130: Would be good to have an explanation what version the user should pick.

bohning: Tried my best to add some sensible info to the tooltip.

RumovZ (approved):

PR #318, opened 2024-11-23
Use different icons for failed downloads and available resources.

bohning:


bohning:
My thinking was that it facilitates the identification of failed downloads (in particular audio/video) and incentivise the average user to at least comment on usdb (or ideally comment a replacement candidate, if available). I was able to identify several of those and fix them on usdb that way.

bohning:
This could also be extended to incate
- when the usdb cover was downloaded as a fallback (metatags url not available)
- when one of the commented resources was downloaded as a fallback
- when the requested container/codec is not available (and a fallback container/codec was downloaded)

bohning:
Superceded by #489 .

PR #324, opened 2024-12-10
feat: Add configuration option whether to trash or delete files/dirs

g3n35i5:
Trashing files obviously doesn't delete them immediately. This causes issues when migrating large USDB databases (thousands of files in the trash bin and no space left on the disk).

As a workaround, I suggest adding an optional environment variable `SYNCER_CLEANUP_DELETE_IMMEDIATELY` which, when set, deletes files and directories immediately instead of trashing them.


RumovZ:
I think if a user feature is relevant enough to be in the app, it should also be discoverable from the GUI. So if you think it is useful to other users, too, I would add a toggle in the settings dialog.

bohning:
Why not have both actions (send to trash / delete permanently) available side by side?

PR #327, opened 2024-12-29
USDB Syncer version annotation updated to latest

Baklap4:
Version wasn't updated within the pyproject.toml which leads local development show 0.7.0 in logs while we're already on 0.9.0

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "poetry.core.masonry.api"
 
 [tool.poetry]
 name = "usdb_syncer"
-version = "0.7.0"
+version = "0.9.0"
 authors = ["Markus Böhning <markus.boehning@gmail.com>"]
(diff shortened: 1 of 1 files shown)


Baklap4:
Not quite sure why the pipeline is failing.. :o

RumovZ:
I wasn't aware we have a version attribute in pyproject.toml. That's an issue, because we release by pushing a version tag, and we should only have a single source of truth for the current version number.
Maybe we should use something like setuptools_scm.

Baklap4:
Hmmm for this pr i could split up the GH runner and yt-dlp update.. so at least those can continue on as the pipeline is currently broken

bohning:
> Maybe we should use something like setuptools_scm.

I think I found a potentially simpler solution using `poetry-dynamic-versioning`, see #329.

PR #329, opened 2025-01-14
Update dependencies.

bohning:
This makes #327 obsolete.


Baklap4:
@bohning Pipeline failing cause of GH runners. Feel free to cherrypick: a1680a8ff7e66fe12cf3ce1a0ee7f56416760869 so we use the 2204 ubuntu image instead of 2404

another way would be to solve the issue

RumovZ (approved):

PR #332, opened 2025-01-22
Add configurable rate limit for yt-dlp.

bohning:
I chose KiB/s as this is what yt-dlp uses in its log messages (and MiB/s). I decided to go for a dropdown field (instead of a spinbox) for ease of use. I am not sure if these are the most sensible options, lower/higher values might be sensible as well.

### DIFF
--- a/src/usdb_syncer/download_options.py
+++ b/src/usdb_syncer/download_options.py
@@ -27,6 +27,7 @@ class AudioOptions:
     bitrate: settings.AudioBitrate
     normalize: bool
     embed_artwork: bool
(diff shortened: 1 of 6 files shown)


Baklap4:
This will download songs on a lower speed thus taking longer. I thought rate-limiting would be more about sleeping for a bit (circuit breaker pattern  `--retry-sleep [TYPE:]EXPR` ) between processing songs so we don't get the message 'please sign in to confirm you're not a bot'. Or is that also solved with lowering the download speed? What are youtube's limits anyways?

RumovZ (approved):

PR #345, opened 2025-02-02
Version check

bohning:


src/usdb_syncer/gui/__init__.py
```diff
@@ -73,6 +80,38 @@ def _run() -> None:
...
 
+def get_latest_version() -> str | None:
+    url = "https://api.github.com/repos/bohning/usdb_syncer/releases/latest"
+    response = requests.get(url, timeout=5)
+    if response.status_code == 200:
+        return response.json()["tag_name"]
+    return None
+
+
+def get_installed_version() -> str | None:
```
RumovZ, line 91: Why not use constants.version?

src/usdb_syncer/gui/__init__.py
```diff
@@ -73,6 +80,38 @@ def _run() -> None:
     app.exec()
 
 
+def get_latest_version() -> str | None:
```
RumovZ, line 83: I would move these functions outside the gui module.

src/usdb_syncer/gui/__init__.py
```diff
@@ -73,6 +80,38 @@ def _run() -> None:
...
+def check_for_update() -> None:
+    latest_version = get_latest_version()
+    installed_version = get_installed_version()
+
+    if latest_version and installed_version:
+        if version.parse(installed_version) < version.parse(latest_version):
+            logging.warning(
+                f"USDB Syncer {latest_version} is available! "
+                f"(You have {installed_version}). Please download the latest release "
+                "from https://github.com/bohning/usdb_syncer/releases/latest."
```
RumovZ, line 107: It might be nice to have a visual indicator that could also be a clickable link somewhere in the GUI. As we know, the log is currently a bit overwhelming.

src/usdb_syncer/gui/__init__.py
bohning: To utils?

bohning: I moved both to utils.

src/usdb_syncer/gui/__init__.py
bohning: I added a hyperlink label next to the search field that is only visible if there is a new version available (and if the app is a bundle).

RumovZ (approved):

Issue #342, opened 2025-02-02
Add option to use the first letter of the artist as a subdirectory


bohning:
A song "Foo - Bar" would then go into path `./F/Foo - Bar/`.

randompersona1:
What do you think about a solution like [this](https://github.com/randompersona1/usdb_syncer/tree/jinja), leveraging jinja? Ignoring the terrible code and the errors for a second, you'd select the first letter with `{{ artist[0] }}/{{ artist }} - {{ title }}`

PR #347, opened 2025-02-03
Use jinja for path templates

randompersona1:
This PR tracks a potential fix for #342 

The basic premise is to use [jinja](https://jinja.palletsprojects.com) for path templates instead of the existing placeholders. The new implementation is:

1. Split the input string into components by "/". Assert there are at least two components.
2. Render each component with jinja.
3. Perform checks on the result of each component, for example for forbidden chars.
4. Run utils.sanitize_filename() over the result of each component.
5. Build the final path

Of course, most of these steps were applied before as well.

Disadvantages:

- new system; incompatible with old placeholders
- new data structure in the settings. For backwards compatibility, I added an empty _tokens list to PathTemplateComponent. Old versions of usdb_syncer will not be able to read the new path template.
- A wiki page explaining some of the features of jinja would probably be necessary

However, jinja has some pretty nice features:

- widespread use in the python ecosystem
- much more powerful templating. Some examples using Foo - Bar from 1985:
  - {{ artist[0] | upper }} / {{ artist }} - {{ title }} / {{ artist }} - {{ title }}: "F/Foo - Bar/Foo - Bar"
  - {{ artist[:2] | upper }} / {{ artist }} - {{ title }} / {{ artist }} - {{ title }}: "FO/Foo - Bar/Foo - Bar"
  - {{ year[-2] ~ "0's" }}/ {{ artist }} - {{ title }} / {{ artist }} - {{ title }}: "80's/Foo - Bar/Foo - Bar"
  - {% if year == 'None' %} Unknown Year {% else %} year {% endif %} / {{ artist }} - {{ title }}: "1985/Foo - Bar"
- though currently not implemented, jinja also supports lists, so you could do some interesting things with the list of languages, like putting all songs *containing* "English" into an english folder and not just the songs

Didn't add custom placeholders yet because I don't understand them.
I didn't bother changing the tests yet, so all of them will fail.

Is this an implementation you would support?


RumovZ:
I'm not sure about this. It's usually better to use an already established solution than implement your own. However, I intentionally didn't use an existing template syntax because they are so much more complex than what we need here.
Maybe we should first check which features the community would actually be using. If it's just the first-letter thing, I would add it to the current implementation. If there are more intricate use cases, I'm open to something like jinja.

randompersona1:
I agree with that. However, I don't need to ask anyone to tell you the placeholders are perfectly fine and it's very easy to add new placeholders for cases like this. I had to reach pretty far to find some interesting examples.

I do still have some thoughts though:
- the implementation I came up with here isn't any more complicated than the existing template implementation imo. Most of the added lines in git are actually from the poetry lockfile, so the changes are truly quite minimal.
- using jinja templates seems equally complex for the end user to the existing templates. Simply swap the colon for two curly brackets.
- changing to jinja templates will break things. So if there was ever a time to do it, it would be pre-1.0.

Like I said though, I'm happy for this not to make it. I simply wanted to throw an alternative (and more general) fix to the linked issue into the room.

Baklap4:
The jinja template looks quite extensive.. not sure if all are needed especially not for the majority of people. Speaking for just myself i could see myself using this over the current implementation and basicly just be done with the templating

> Maybe we should first check which features the community would actually be using

sounds like a good plan -> Poll on discord?

randompersona1:
Since there doesn't seem to be a need or interest, I'm closing.

PR #349, opened 2025-02-09
Support zipped addons

randompersona1:
The syncer should import addons from zip files to enable easier distribution. Python supports this natively, so all we have to do is add the zip file to `sys.path`. The old behavior of importing all modules stays.

To create a zipped addon, simply zip the module folder, so that the following file structure is output:
```
.
└── addon.zip/
    └── addon/
        ├── __init__.py
        ├── file1.py
        └── file2.py
```

### DIFF
--- a/src/usdb_syncer/addons.py
+++ b/src/usdb_syncer/addons.py
@@ -9,7 +9,15 @@ from usdb_syncer.logger import logger
 
 
 def load_all() -> None:
-    sys.path.append(str(utils.AppPaths.addons))
-    for _, mod, is_pkg in pkgutil.iter_modules([str(utils.AppPaths.addons)]):
+    addons_path = utils.AppPaths.addons
+    sys.path.append(str(addons_path))
+
+    # Search for zip files in the addons directory and add them to the path
+    zip_files = [str(z) for z in addons_path.glob("*.zip")]
+    for zip_file in zip_files:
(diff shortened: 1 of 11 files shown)


src/usdb_syncer/addons.py
```diff
@@ -9,7 +9,17 @@
 
 
 def load_all() -> None:
-    sys.path.append(str(utils.AppPaths.addons))
-    for _, mod, is_pkg in pkgutil.iter_modules([str(utils.AppPaths.addons)]):
+    addons_path = utils.AppPaths.addons
+    sys.path.append(addons_path.as_posix())
```
RumovZ, line 13: Though modern Windows systems support posix paths, one should actually use str(path).

src/usdb_syncer/addons.py
```diff
@@ -9,7 +9,17 @@
...
+
+    # Search for zip files in the addons directory and add them to the path
+    zip_files = list(addons_path.glob("*.zip"))
```
RumovZ, line 16: Can also make strs directly: str(z) for z in addons_path.glob("*.zip")

randompersona1:
Cleaned it up. Do we explain this in the addon readme?

RumovZ:
Yeah, good idea.

RumovZ (approved):
Thanks!

PR #350, opened 2025-02-15
Log local songs that long longer exist on usdb.

bohning:
Maybe it makes sense to trash these songs as well (e.g. deleted duplicates), but maybe that should be configurable in the settings.


RumovZ:
TODO: remove folder when downloading not existing song.

RumovZ (approved):

PR #358, opened 2025-02-23
Add additional browsers supported by rookiepy.

bohning:


bohning:
This PR seems to be problematic as the browser values have changed (capitalization), see https://github.com/bohning/usdb_syncer/issues/365. Not sure how that situation can be solved.

PR #359, opened 2025-02-23
Fix image processing

bohning:
This fixes multiple issues:
- non-jpg image formats (png, webp) were erroneously saved with .jpg extension, which turned out to be a problem with MelodyMania.
- different order of operations for crop and resize (cover: first crop, then resize - background: first resize, then crop) is now in place
- image files with CMYK color channels are now correctly converted to RGB color channels before saving (which was only an issue if no post-processing took place, which always triggered the RGB conversion)

### DIFF
--- a/src/usdb_syncer/resource_dl.py
+++ b/src/usdb_syncer/resource_dl.py
@@ -4,7 +4,7 @@ import io
 import os
 from enum import Enum
 from pathlib import Path
-from typing import Union, assert_never
+from typing import Callable, Union, assert_never
 
 import filetype
 import requests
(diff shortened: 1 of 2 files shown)


src/usdb_syncer/resource_dl.py
```diff
@@ -215,33 +216,54 @@ def download_and_process_image(
...
-                    (resize.width, resize.height), resample=Resampling.LANCZOS
-                )
-            if meta_tags.contrast == "auto":
-                image = ImageOps.autocontrast(image, cutoff=5)
-            elif meta_tags.contrast:
-                image = ImageEnhance.Contrast(image).enhance(meta_tags.contrast)
+            operations = {
+                ImageKind.COVER: ["rotate", "crop", "resize", "contrast"],
+                ImageKind.BACKGROUND: ["resize", "crop"],
+            }.get(kind, ["rotate", "crop", "resize", "contrast"])
```
RumovZ, line 240: Defining and matching magic strings is error prone. Instead you could factor out the processing steps, define a list of these functions according to image kind, and then call them in a loop.

bohning: Like this?

src/usdb_syncer/resource_dl.py
```diff
@@ -215,33 +216,63 @@ def download_and_process_image(
...
-                image = image.crop((crop.left, crop.upper, crop.right, crop.lower))
-            if resize := meta_tags.resize:
-                image = image.resize(
-                    (resize.width, resize.height), resample=Resampling.LANCZOS
-                )
-            if meta_tags.contrast == "auto":
-                image = ImageOps.autocontrast(image, cutoff=5)
-            elif meta_tags.contrast:
-                image = ImageEnhance.Contrast(image).enhance(meta_tags.contrast)
+            for process in IMAGE_PROCESSING_STEPS.get(
```
RumovZ, line 271: Looks good! Just one more lint: imagine there was a new image kind. You'd have to remember to add it to the dict as well. That's why I'd use an exhaustive match expression instead.

bohning: Thanks for the value feedback. Like this? This actually simplifies it.

RumovZ: Yeah, great!

RumovZ (approved):

PR #360, opened 2025-03-01
Enhance resource error handling

bohning:


src/usdb_syncer/discord.py
```diff
@@ -0,0 +1,33 @@
...
+
+import discord
+
+from usdb_syncer import SongId
+from usdb_syncer.constants import Usdb
+from usdb_syncer.usdb_song import UsdbSong
+
+BASE_URL = "https://discordapp.com/api/webhooks"
+CHANNEL = "1231022685513973840"
+HASH = "0MfBIqFH9JFl4zALvqdm-etzIuT6OefxH12YXjCTMw7xyyhUlSL1oJkxjnifQmrZ2tzD"
```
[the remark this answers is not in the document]
randompersona1: I don't think spam is a big issue. It's not possible to hide the url from a malicous actor, and worst case, the feature is useless. The benefits outweigh here IMO.

Agreed on the issue of updating the URL. I don't see a simple way of solving though. Perhaps a static web page on GitHub Pages?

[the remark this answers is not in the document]
bohning: Couldn't we put the URL in a file on github and send a request to fetch it e.g. on startup of the Syncer? This way, it could be changed later on.

pyproject.toml
```diff
@@ -33,6 +33,7 @@ setuptools = "^75.8.0"
 unidecode = "*"
 yt-dlp = {version = "*", extras = ["default", "curl_cffi"]}
 wheel = "*"
+discord = "^2.3.2"
```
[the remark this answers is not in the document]
bohning: Good point, I will look into it. It's probably easily solvable via requests.

src/usdb_syncer/resource_dl.py
```diff
@@ -151,18 +154,37 @@ def _download_resource(options: YtdlOptions, resource: str, logger: Log) -> str
...
+                        "Countries, where the resource is available: "
+                        + ", ".join(utils.get_allowed_countries(resource))
+                    )
                 return None
+            if YtErrorMsg.YT_NOT_AVAILABLE in str(e):
+                logger.warning(
+                    "Resource no longer available. Help the community, find a suitable "
+                    "replacement and comment it on USDB."
+                )
+                notify_discord(cast(SongLogger, logger).song_id, url)
```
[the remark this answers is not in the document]
bohning: Not sure how to address this. Any hints? How should I get the song_id instead?

bohning:
> If we decide to implement this, it should probably be opt-in or users might negatively view it as some kind of telemetry.

I added a setting to disable anonymous Discord notifications.

src/usdb_syncer/discord.py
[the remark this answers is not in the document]
RumovZ: That sounds very useful! Shall we make it more generic so we can use it for all future live parameters as well? Like a single config.json that that gets fetched on startup.
Would be great for reviewing to have it in a separate PR, though. Once it's merged, it should be simple to update this one to use the new feature.

src/usdb_syncer/discord.py
```diff
@@ -0,0 +1,44 @@
+"""Functions for Disord integration."""
+
+import requests
+
+from usdb_syncer import SongId, remote_config
+from usdb_syncer.constants import Usdb
+from usdb_syncer.logger import Log
+from usdb_syncer.usdb_song import UsdbSong
+
+sent_notifications: set[tuple[SongId, str]] = set()
```
RumovZ, line 10: This should be persisted in the db.

src/usdb_syncer/gui/forms/SettingsDialog.ui
```diff
@@ -543,6 +553,41 @@
...
+            <item row="0" column="1">
+             <widget class="QCheckBox" name="checkBox_discord_allowed">
+              <property name="toolTip">
+               <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;This will send a notification to our Karaoke Discord channel and report unavailable resources so that they can be fixed. This reporting is completely &lt;span style=&quot; font-weight:700; font-style:italic;&quot;&gt;anonymous&lt;/span&gt;. Please keep this enabled to support the community.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
+              </property>
+              <property name="text">
+               <string/>
+              </property>
+              <property name="checked">
+               <bool>true</bool>
```
RumovZ, line 584: Would really prefer to have this opt-in. Surely, there would still be more than enough participants.

src/usdb_syncer/resource_dl.py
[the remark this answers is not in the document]
bohning: I've refactored some more and moved the notify_discord to the higher-level functions.

src/usdb_syncer/resource_dl.py
```diff
@@ -29,6 +32,40 @@
 YtdlOptions = dict[str, Union[str, bool, tuple, list, int]]
 
 
+class ResourceInvalidError(Exception):
```
RumovZ, line 35: There is a common base class in errors.py.

src/usdb_syncer/resource_dl.py
```diff
@@ -119,14 +152,34 @@ def download_video(
...
     ydl_opts = _ytdl_options(
         options.ytdl_format(), browser, path_stem, options.rate_limit
     )
-    if filename := _download_resource(ydl_opts, resource, logger):
-        return os.path.splitext(filename)[1][1:]
-    return None
+    return _download(resource, ydl_opts, logger)
+
+
+def _download(resource: str, options: YtdlOptions, logger: Log) -> ResourceDLResult:
```
RumovZ, line 163: It's great that now a type-safe result object is returned!
However, I don't understand the necessity of creating an exception in `_download_resource` that is then immediately replaced by an enum variant in `_download`. Why not directly return the `ResourceDLError` (or `ResourceDLResult`)?
Also, logging these exception is not helpful if they don't have a custom `__str__` method or similar.

bohning: Thanks for the detailed review and your comments, really appreciated! I think I've considered them all and changed the code accordingly.

RumovZ: Looks great! The exception classes can be removed now, right?

bohning: Yes, I forgot to remove them. Would have thought that pylint/mypy would have pointed out unused classes... They're removed now.

CHANGELOG.md
```diff
@@ -3,9 +3,13 @@
 # Changes
 
 ## Fixes
+- Audio/video format selection regression in 0.11.0 fixed
+- Download of non-jpg images are now correctly converted to jpg.
+- Image processing order is now correctly handled (crop before resize for covers, resize before crop for backgrounds)
 
 ## Features
 - The Syncer now checks if a new version is available.
+- Info about unavailable or invalid resources are now sent to Discord (iplease enable this in the settings to support the community)
```
RumovZ, line 12:
```suggestion
- Audio/video format selection regression in 0.11.0 fixed.
- Downloads of non-jpg images are now correctly converted to jpg.
- Image processing order is now correctly handled (crop before resize for covers, resize before crop for backgrounds).

## Features
- The Syncer now checks if a new version is available.
- Infos about unavailable or invalid resources are now sent to Discord (please enable this in the settings to support the community).
```

src/usdb_syncer/resource_dl.py
RumovZ: I think it only works for private names as public ones could be part of a public API.

RumovZ (approved):

PR #364, opened 2025-03-03
Fix linux execution and builds

randompersona1:
ubuntu-22.04 uses glibc 2.35 instead of 2.39 for ubuntu-latest (24.04). Increases compatibility with systems on older glibc versions.


RumovZ (changes requested):
Why not an even earlier image like 20.04? Did you try? I guess at some point you won't be able to get Python 3.12 anymore.

In any case, it is the runner of the build job that has to be changed (and probably that of the tox job).

randompersona1:
Because the readme states that we support glibc 2.35 specifically. I wasn't sure if there was a reason for that and decided to merely restore to that instead of testing older versions. I'll spin up a VM and check. 

My bad for the wrong change. Didn't realize that was how matrixes worked.

RumovZ:
Yeah, I don't really think that section in the readme is accurate. Let's fix that, too, after we've figured out the best runner.
You should be able to test compatibility by changing the runner of the tox job and pushing. At least, that job passing means Python 3.12 can be installed on that runner.

randompersona1:
Alright. [I tested with 20.04 and it builds](https://github.com/randompersona1/some_actions_testing/actions/runs/13750901040), although I haven't had the chance to run it. Definitely want to do that before finalizing the changes. But it seems as though there are no issues on that front.

However, I also discovered that [the 20.04 runner is being depreciated](https://github.com/actions/runner-images/issues/11101). That wouldn't be an issue in itself, since we could just use 22.04, but I have no desire to figure things out when github decides to depreciate _that_, which will likely happen in less than two years.

Instead, we could use a docker image for building the linux releases. In particular, [manylinux](https://github.com/pypa/manylinux) provides neat images with specific dependancies (including glibc). I tested this with `manylinux2014`, which is based on `glibc 2.17`, and the build works fine, though again, I haven't tried running. This would allow us to build for the maximum number of linux users, while keeping development work fairly low. I am strongly in favor of this, but of course simply building with `ubuntu-22.04` also works.

RumovZ:
Seems like using a docker image for builds gives better compatibility and reproducibility. I like it!

randompersona1:
I'm fairly certain I have a working CI over at https://github.com/randompersona1/some_actions_testing/blob/master/.github/workflows/release.yaml. I'll put it in this PR soon, but I can freely test in that repository (which is necessary when looking at the amount of actions I've used).

Here's the rough yet hopefully complete list of changes I made so far.
- Changed the runner to `quay.io/pypa/manylinux_2_34_x86_64`. This means that `glibc` is guaranteed to be compatible with everything >=2.34. From my testing, it only actually linked to 2.27, which is even better compatibility. 
Unfortunately, I had to resort to building python from source for this. The python binaries included by manylinux are compiled without a certain flag (https://github.com/pypa/manylinux/issues/1149). It's less horrible than it sounds, because python doesn't require all that many dependancies, but building from source does take ~15min (sorry github). Don't see another way though.
- Removed installing Qt? I dont't understand why it was there in the first place, but both Windows and Linux builds work perfectly without. Is this a mac thing?
- Removed all occurrences of dbus-python, because it's not actually needed. Modern linux keyring backends are compatible with keyrings `SecretStorage`, which doesn't need dbus-python.
- Changed `src/usdb_syncer/db/__init__.py::delete_resource_files`: The sql syntax with `IN` wasn't compatible with the version of sqlite that was linked into the binary, causing an error when a resource file was deleted. I changed it to use a normal `WHERE` syntax, and haven't observed any issues (though I literally haven't even checked if it works, just that it doesn't error like before). On that note, there might be more similar issues with sqlite in particular. I haven't tested much functionality of the syncer on Linux, just stumbled over that one in particular.
- Temporarily changed the `release.yaml` so that I don't have to tag every commit by replacing ${{ github.refs }} with a hardcoded 0.11.0. Also removed the release workflow for brevity.

A complete build for everything with release artifacts is here: https://github.com/randompersona1/some_actions_testing/actions/runs/13798616593

So for a ToDo:
- test all aspects of the linux version
- check the dependancies of the python build because I'm fairly sure most of the graphics stuff isn't actually needed
- update the readme to reflect linux changes
- restore the release.yaml to full functionality

RumovZ:
Oh, looks to require a lot of effort after all. Thanks for sticking with it!

> but building from source does take ~15min

That's painful, but may be acceptable as it only affects the release pipeline.

> Removed installing Qt? I dont't understand why it was there in the first place, but both Windows and Linux builds work perfectly without. Is this a mac thing?

No, it's for Linux. We don't use it to install Qt itself, but its dependencies. You may be able to run a bundle without it, but some features like playback won't work then. That may depend on the system it's running on, so it's hard to test.
But we cannot just remove it unless we're sure the necessary dependencies are still bundled.

> Removed all occurrences of dbus-python, because it's not actually needed. 

Same here. May work on your system, but I'm skeptical that's universally true. There was a reason we added it.

Your Ubuntu package is 50 MB smaller. That means either we currently include a lot of bloat or you're missing some important dependencies—or both. 🙂

randompersona1:
I've tested fairly extensively on every modern distro I could think of, and almost everything works perfectly. I have discovered two errors though:

1. Opening a song in the browser fails. This is a regression from the current binaries, and I really don't understand how. The webbrowser module doesn't provide much debug info either, so I'm a bit lost here.
2. Opening a song in the file browser (sometimes) fails with an exit code of four. Luckily, this happens with the current binaries too 😃, but not when running from source. Also very confusing.

Other than that, the critical functions are good. Playback works, downloads work, keyring works. Builds hopefully work as well. Still don't know why the binary is smaller.

randompersona1:
In addition to the build changes I've fixed a couple of linux-specific bugs, both already existing and newly introduced. Particularly notable is the new `LinuxEnvCleaner` context manager: it needs to be used before accessing specific system libraries. Unfortunately, this is impossible to test without actually creating and testing a linux binary.

bohning:
Did you have a chance to look at this @RumovZ ? Can this be merged when the conflicts in `poetry.lock` are resolved?

RumovZ:
No, I'm just returning from abroad. Will try to have a look the next couple of days.

README.md
```diff
@@ -31,21 +31,6 @@ git clone https://github.com/bohning/usdb_syncer.git
...
 
-<details>
-
-<summary>If you're on <b>Linux</b>, make sure required packages are installed.</summary>
-
-```bash
-apt install -y gcc python3-dev libdbus-1-dev
-pkg-config --cflags --libs dbus-1
-# required by Qt
-apt install -y build-essential libgl1-mesa-dev libgstreamer-gl1.0-0 libpulse-dev libxcb-glx0 libxcb-icccm4 libxcb-image0 libxcb-keysyms1 libxcb-randr0 libxcb-render-util0 libxcb-render0 libxcb-shape0 libxcb-shm0 libxcb-sync1 libxcb-util1 libxcb-xfixes0 libxcb-xinerama0 libxcb1 libxkbcommon-dev libxkbcommon-x11-0 libxcb-xkb-dev
```
RumovZ, line 42: I tried with WSL Ubuntu 20.04 and Qt won't install without these (and additionally libxcb-cursor0; multimedia still not working).

README.md
```diff
@@ -105,58 +90,52 @@ startup phase, they cannot be completely ruled out.
 
 ## Troubleshooting
 
-- may require extra packages on Linux
-
-  ```bash
-  apt update
-  apt install libdbus-1-3
```
RumovZ, line 112: I needed to install gnome-keyring, but you seem to be right that we don't really need dbus, which is neat because it makes things a lot simpler!

README.md
```diff
@@ -105,58 +90,52 @@ startup phase, they cannot be completely ruled out.
...
-  :white_check_mark: should work
-
-confirmed support:
-
-- Ubuntu 23.04
+Linux builds are generated on AlmaLinux 9. They should be compatible with any modern distribution. If not, please open an issue.
+
+The only known requirement is `glibc >= 2.34`. The following table (based on <https://pkgs.org/search/?q=glibc>, 8.7.2023) summarizes different Linux distributions for having greater or equal version of `glibc`. For (likely) supported distributions the minimum OS Version is given that has a required glibc version. For (likely) unsupported distributions the recent highest Versions (if known) of Linux Distributions with its highest glibc version is given:
+
+|                    |         OS          |    OS Version    | glibc |
```
RumovZ, line 110: Is it worth it maintaining this table or should we just get rid of it? I gather the glibc requirement says it all essentially.

src/usdb_syncer/utils.py
```diff
@@ -10,6 +10,7 @@
 import time
 import unicodedata
 from pathlib import Path
+from typing import Any, Optional, Type
```
RumovZ, line 13: Modern annotations use `|` and `type` instead.

RumovZ (commented):
Another issue with old sqlite versions: `sqlite_schema` used to be named `sqlite_master` (https://sqlite.org/schematab.html). Would be nice to make this version-agnostic as well while you're already on it.

@bohning, to minimize merge conflicts, it would help to only update dependencies if necessary, and outside long-living unrelated PRs.

README.md
randompersona1: Ah, I messed this up in git. Those are needed of course.

I didn't see the multimedia errors, but I barely tested running from source and focused on the binary. Is ffmpeg installed?

randompersona1: Testing WSL for the first time, we also require `libva-dev libva-drm2 libva-x11-2 libpipewire-0.3-dev`. I suspect libpipewire is WSL-specific, while libva is probably needed on all distros. I didn't catch this earlier because I installed ffmpeg on my virtual machines, and the apt ffmpeg package relies on those libva libraries and installed them. I'll add those to the readme.

README.md
randompersona1: I replaced it with a short instruction on how to check the current glibc version.

README.md
RumovZ: Had no luck getting multimedia to work, yet. libpipewire-0.3-0 doesn't seem to be available for Ubuntu 20.04 which might be the reason.
Let's leave it at that for now.

bohning: I got it to work on a fresh Ubuntu 20.04 (as VM) after a bit of fiddling. After adding `sudo add-apt-repository ppa:pipewire-debian/pipewire-upstream`, I was able to install `libpipwire-0.3-0`. The other thing in the way was that Qt ≥ 6.5 requires openssl 3 instead of openssl 1. So I basically followed [these steps](https://maggiminutes.com/install-latest-openssl-3-on-linux/) to compile and install openssl 3 (confirmed with `openssl version`).

When I start the syncer, I still see
```
qt.multimedia.symbolsresolver: Couldn't resolve va symbol vaSyncBuffer
qt.multimedia.symbolsresolver: Couldn't resolve va symbols
```
but it doesn't seem to affect sample / audio playback.

randompersona1: I think Ubuntu 20.04 is too old to be relevant and too annoying to fix. Ubuntu 22.04 works fine (and doesn't even need libpipewire for some reason). Running on 20.04 still works because simple audio playback doesn't require any of this, the user would just get the warnings.

randompersona1:
I've updated the list of required dependancies in the readme. It's hard to trace when and why each one was added, but I've tested that the syncer runs on both Ubuntu22.04 and Debian 12.

RumovZ (approved):
Thanks a lot for your persistence! Bundling truly is a pain.
Looking forward to try the next release!

PR #367, opened 2025-03-09
Improve handling of artist initial for path templates.

bohning:
The first alphanumerical character is decoded to it's closest ASCII representation, which is then uppercased.


bohning:
How shall we handle cases where the unidecode result contains more than one character? E.g. `Ю` is unidecoded to `IU`.

RumovZ:
Just using the first I guess.

RumovZ (approved):

PR #369, opened 2025-03-16
Add option to write replay gain headers for loudness normalization.

bohning:
With https://github.com/UltraStar-Deluxe/USDX/pull/936 on the way, this adds the option to write replay gain headers for loudness normalization as an alternative to rewriting the entire file, which is the more desirable approach. However, support by Karaoke software so far is only in the pipeline for UltraStar deluxe with above PR, but maybe others will follow.

Although it would be better to use the normalization stats from a 2-pass run, `ffmpeg-normalize` currently doesn't offer this possibility directly (however, workarounds exist, see https://github.com/slhck/ffmpeg-normalize/issues/278). Potentially, `ffmpeg-normalize` integrates writing replay gain headers directly in a future version, we'll have to see.


complexlogic:
Thanks a lot for implementing this feature in the Syncer. After https://github.com/slhck/ffmpeg-normalize/commit/e1ec6713cbe6c291c9f71de9b6d896cb46a95468 is included in a release, would you consider using 2 pass results for the ReplayGain tags?

bohning:
Yes, I am aiming for using the 2-pass results. From the last sentence in the owners [answer](https://github.com/slhck/ffmpeg-normalize/issues/278#issuecomment-2725606339), I read that he maybe plans to add the option to write replay gain headers directly using ffmpeg-normalize - or maybe I am reading that wrong?

complexlogic:
Maybe. It looks like ffmpeg-normalize doesn't currently support metadata writing. So to implement that feature, the developer would likely need to add another dependency to the project (probably mutagen) and determine whether that impact justifies the value of the new feature.

Audio metadata is not as straightforward as it seems. There are many small details that can be easily overlooked, and then cause compatibility issues between different software. In the case of ReplayGain, we previously discussed the uppercase vs. lowercase issue, and the Opus issue.

I believe it is valuable to be able to control the tag writing from within the Syncer, instead of relying on an external implementation, due to the issues noted above. You have already done the work of writing the code, and testing it to ensure it will work nicely with USDX. So why not use it?

For these reasons, my recommendation is to keep your current tag writing code. If you instead choose to let ffmpeg-normalize write the tags, we should work closely with the developer to ensure the implementation is correct, so it will work properly with USDX.

bohning:
Okay, convincing argument.

complexlogic:
I just pulled the latest commits. Everything is working as expected. The gain values are now accurate, and the tag formatting is correct. I tested the files with https://github.com/UltraStar-Deluxe/USDX/pull/936 and verified that the ReplayGain tags are successfully read.

I recommend merging this feature.

CHANGELOG.md
```diff
@@ -9,7 +9,10 @@
 
 ## Features
 - The Syncer now checks if a new version is available.
-- Infos about unavailable or invalid resources are now sent to Discord (please enable this in the settings to support the community).
+- Infos about unavailable or invalid resources are now sent to Discord (**please enable this in the settings to support the community**).
+- Songs can now be rated on USDB via the Syncer (1-5 star rating).
+- Artist initial is now available for templates.
+- Audio normalization has a new option ReplayGain. If your karaoke software supports ReplayGain, this is the preferred option as it does not reencode the audio file but instead only write normalization information into the header of the audio file.
```
RumovZ, line 15:
```suggestion
- Audio normalization has a new option ReplayGain. If your karaoke software supports ReplayGain, this is the preferred option as it does not reencode the audio file but instead only writes normalization information into the header of the audio file. 
```

src/usdb_syncer/constants.py
```diff
@@ -1,11 +1,13 @@
 """Constants"""
 
 import re
+import sys
 
 # set in the release workflow
 VERSION = "dev"
 COMMIT_HASH = "dev"
 SHORT_COMMIT_HASH = COMMIT_HASH[:7]
+NUL = "NUL" if sys.platform == "win32" else "/dev/null"
```
RumovZ, line 10: Does not fit the block's comment. Also the variable name could be a bit more descriptive.

src/usdb_syncer/resource_dl.py
```diff
@@ -30,6 +41,11 @@
     )
 }
 
+DEFAULT_TARGET_LEVEL_RG = -18.0  # dB
+DEFAULT_TARGET_LEVEL_R128 = -23.0  # dB
```
RumovZ, line 45: Might be more helpful to include units in the variable names.

RumovZ (commented):
Should we maybe add a postprocessing module for the related logic currently spread across song_loader.py and resource_dl.py?

src/usdb_syncer/constants.py
bohning: Just found out that there is `os.devnull`, which makes things cleaner.

RumovZ (commented):
`context` is an extremely generic module name. You wonder, context for what — apparently the song loader. But then it should really be in the according module.
I assume you moved it so it can be passed to the postprocessing - functions. But you could just pass the necessary attributes of Context, like Txt.
This would also make the module interface clearer. Passing a huge mutable object like Context through several modules makes it very hard to understand, where the final values are coming from.

RumovZ (approved):

PR #372, opened 2025-03-24
Add hook for main window load

randompersona1:
This hook is called when the main window has loaded and passes the MainWindow object to subscribers. This allows them to modify the main window if needed.

### DIFF
--- a/src/usdb_syncer/hooks.py
+++ b/src/usdb_syncer/hooks.py
@@ -24,7 +24,7 @@ P = ParamSpec("P")
 class _Hook(Generic[P]):
     """Base class for hooks."""
 
-    _subscribers: list[Callable[P, None]] = attrs.field(factory=list)
+    _subscribers: list[Callable[P, None]]
 
     def __init_subclass__(cls) -> None:
         cls._subscribers = []
@@ -43,7 +43,8 @@ class _Hook(Generic[P]):
             try:
                 func(*args, **kwargs)
             except Exception as e:  # pylint: disable=broad-except
-                logger.warning(f"Plugin error in {func.__name__}: {e}")
+                logger.debug(e, exc_info=True)
+                logger.warning(f"Plugin error in {func.__name__}: {e.__class__.__name__}")
 
 
 class SongLoaderDidFinish(_Hook):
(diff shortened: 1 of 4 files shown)


src/usdb_syncer/hooks.py
```diff
@@ -37,7 +40,10 @@ def unsubscribe(cls, func: Callable[P, None]) -> None:
     @classmethod
     def call(cls, *args: P.args, **kwargs: P.kwargs) -> None:
         for func in cls._subscribers:
-            func(*args, **kwargs)
+            try:
+                func(*args, **kwargs)
+            except Exception as e:  # pylint: disable=broad-except
+                logger.warning(f"Plugin error in {func.__name__}: {e}")
```
RumovZ, line 46: Might to be nice to have the traceback, at least in the debug log.

src/usdb_syncer/hooks.py
```diff
@@ -21,7 +24,7 @@
 class _Hook(Generic[P]):
     """Base class for hooks."""
 
-    _subscribers: list[Callable[P, None]]
+    _subscribers: list[Callable[P, None]] = attrs.field(factory=list)
```
RumovZ, line 27: Any specific reason for adding this? This field is only relevant for subclasses for which it is set in `__init_subclass__`.

randompersona1: Not really, which is to say, not at all. I saw it online.

src/usdb_syncer/hooks.py
randompersona1: Good point.

src/usdb_syncer/hooks.py
```diff
@@ -37,7 +40,13 @@ def unsubscribe(cls, func: Callable[P, None]) -> None:
     @classmethod
     def call(cls, *args: P.args, **kwargs: P.kwargs) -> None:
         for func in cls._subscribers:
-            func(*args, **kwargs)
+            try:
+                func(*args, **kwargs)
+            except Exception as e:  # pylint: disable=broad-except
+                logger.debug(e, exc_info=True)
```
RumovZ, line 46: This logs the exception message twice.

src/usdb_syncer/hooks.py
```diff
@@ -37,7 +40,13 @@ def unsubscribe(cls, func: Callable[P, None]) -> None:
...
+                logger.warning(
+                    f"Plugin error in {func.__name__}: {e.__class__.__name__}"
```
RumovZ, line 48: Here, I would think the exception message is more friendly / useful than its class name.

RumovZ (approved):
Thanks!

PR #376, opened 2025-03-27
Overhaul cookies and auth

randompersona1:
I've switched some things around and (hopefully) made everything regarding cookies and auth more concise. Also got rid of those horrible tempfiles we used to pass around cookies.

I still think the UX could use some improvement. I'm not quite sure what the expected behaviour is when the browser selector is used in the usdb login screen? There should also be some more carefully placed debug and info logs, but I'd prefer some feedback first.


randompersona1:
I've streamlined the use of cookies massively here. They're now properly passed through the download options to allow lower download functions to use them instead of accessing settings. I've also improved naming, and some other details.

The new `authentication` module handles all of the backend, providing nice functions that hide the use of files and encryption. The settings seemed too cluttered for this.

I've also slightly modified the settings dialog for some better UX, and to check whether the saved cookies still exist. I still think the USDB login dialog should have the same "from file" option as in the settings, but that can be done in the main branch.

PR #379, opened 2025-04-04
Warn if txt file contains misused #VIDEOGAP.

bohning:
#VIDEOGAP aligns audio and video from different sources (e.g. a live video to an album version audio), so both `a=` and `v=` metatags need to be present. In the case of audio only, this has no negative side effects (as there is no video to align to), but in case of audio and video from the same source, a #VIDEOGAP misaligns the video’s audio. Therefore, it makes sense to log this and remove the #VIDEOGAP from the local file.


src/usdb_syncer/song_txt/__init__.py
```diff
@@ -49,6 +49,19 @@ def parse(cls, value: str, logger: Log) -> SongTxt:
         lines = [line for line in value.splitlines() if line]
         headers = Headers.parse(lines, logger)
         meta_tags = MetaTags.parse(headers.video or "", logger)
+        if headers.videogap is not None:
```
RumovZ, line 52: This would be better placed in some sanitation code than in the parse code.

bohning: Ah, good point! I moved it to a `fix_videogap()` sanitation function.

RumovZ (approved):

PR #381, opened 2025-04-08
Handle unavailable format gracefully.

bohning:
Less common sites may offer limited audio/video formats, so the requested format might not be available. We can fallback to whatever best format is available. At least ffmpeg-based Karaoke softwares should be able to handle the resulting file (fixes e.g. https://usdb.animux.de/index.php?link=detail&id=14856).

### DIFF
--- a/src/usdb_syncer/constants.py
+++ b/src/usdb_syncer/constants.py
@@ -115,6 +115,7 @@ class YtErrorMsg:
     YT_UNAVAILABLE = "Video unavailable"
     YT_PARSE_ERROR = "Failed to parse XML"
     YT_FORBIDDEN = "HTTP Error 403: Forbidden"
+    YT_FORMAT_NOT_AVAILABLE = "Requested format is not available."
 
 
 SUPPORTED_VIDEO_SOURCES_REGEX = re.compile(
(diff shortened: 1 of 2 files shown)


bohning:
> Might the user not prefer to skip downloading entirely than having an unusable song?

He might, he might not. The song (most likely) just works (e.g. USDX or any other ffmpeg-based karaoke software), he might want to manually convert the song (and be happy to have something to work with), he might have a [reencoder](https://gist.github.com/randompersona1/da1cd35969e9f585777744576c617c11) addon in place that does this for him.

RumovZ:
> he might want to manually convert the song (and be happy to have something to work with)

Then they should configure their settings accordingly.
I also find it confusing code-wise. There is some intricate logic to build the format specifier. But with this PR it just gets overwritten in the end.

bohning:
I understand your concern. What if the user generally wants mp4 (and not "best available") except for those rare cases where mp4 is not available, where he just takes whatever there else is?

RumovZ:
I generally think "almost always" is the wrong behaviour for any software to have.
If the only reason a user would want the file anyway is to reencode it later, then shouldn't the syncer do it for them?

Also note that the implementation in this PR disregards resolution and FPS settings.

bohning:
True, then we need to add video encoding functionality. Disregarding resolution and fps settings was indeed on purpose here because the available formats for https://usdb.animux.de/index.php?link=detail&id=14856 simply do not offer this information. I am open to any other solution for 14856 that gracefully handles the case that the requested format is not available.

PR #383, opened 2025-04-09
Fix pdf

bohning:


src/usdb_syncer/pdf.py
```diff
@@ -1,40 +1,217 @@
...
+            case _:
+                pass
+    entry += "</font>"
+    return entry
+
+
+def _add_page_number(canvas: Canvas, doc: BaseDocTemplate) -> None:
+    canvas.saveState()
+    page_num: str = str(doc.page)
+    canvas.setFont("NotoSans-Regular", 8)
```
RumovZ, line 199: The font names are repeated multiple times and should be constants.

src/usdb_syncer/gui/forms/ReportDialog.ui
```diff
@@ -0,0 +1,318 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
RumovZ, line 1: I think the source selector should be available across all report kinds, and Ok should generate the report. You'd still have the tab widget to choose the report kind and specific settings. You'd also need a Cancel button.

At this point you can also add a check if the selector matches any songs at all.

src/usdb_syncer/pdf.py
```diff
@@ -1,40 +1,217 @@
...
+from reportlab.pdfgen.canvas import Canvas
+from reportlab.platypus import BaseDocTemplate, Flowable, Frame, PageTemplate, Paragraph
 
 from usdb_syncer import SongId
+from usdb_syncer.gui.song_table.column import Column
+from usdb_syncer.settings import ReportOrientation, ReportPagesize
 from usdb_syncer.usdb_song import UsdbSong
 
+pdfmetrics.registerFont(
+    TTFont("NotoSans-Black", "./src/usdb_syncer/gui/resources/NotoSans-Black.ttf")
```
RumovZ, line 22: Paths should not be hardcoded strings but Path objects in AppPaths (at least the parent folder).

src/usdb_syncer/pdf.py
```diff
@@ -1,40 +1,217 @@
...
+        case _ as unreachable:
+            assert_never(unreachable)
+    return (
+        landscape(pagesize) if orientation == ReportOrientation.LANDSCAPE else pagesize
     )
 
+
+def _create_paragraph_styles(base_font_size: int) -> dict[str, ParagraphStyle]:
+    return {
+        "Initial": ParagraphStyle(
```
RumovZ, line 106: Should be type-safe, e.g. by using a dataclass instead of a dict with magic keys.

RumovZ (changes requested):
Unfortunately, it doesn't work for me. Here is the traceback after propogating the exception (see below):

```
2025-04-12 09:18:01 [ERROR] Uncaught exception:
    Traceback (most recent call last):
      File "C:\code\Python\usdb_syncer\src\usdb_syncer\gui\progress.py", line 87, in wrapped_on_done
    on_done(result)
      File "C:\code\Python\usdb_syncer\src\usdb_syncer\gui\report_dialog.py", line 109, in on_done
    result.result()
      File "C:\code\Python\usdb_syncer\src\usdb_syncer\gui\progress.py", line 36, in result
    raise self._result.error
      File "C:\code\Python\usdb_syncer\src\usdb_syncer\gui\progress.py", line 79, in wrapped_task
    result = Result(task())
                    ^^^^^^
      File "C:\code\Python\usdb_syncer\src\usdb_syncer\gui\report_dialog.py", line 139, in <lambda>
    lambda: generate_report_pdf(
            ^^^^^^^^^^^^^^^^^^^^
      File "C:\code\Python\usdb_syncer\src\usdb_syncer\pdf.py", line 80, in generate_report_pdf
    doc.build(content)
      File "~\AppData\Local\pypoetry\Cache\virtualenvs\usdb-syncer-VPHy5iGw-py3.12\Lib\site-packages\reportlab\platypus\doctemplate.py", line 1109, in build
    self._endBuild()
      File "~\AppData\Local\pypoetry\Cache\virtualenvs\usdb-syncer-VPHy5iGw-py3.12\Lib\site-packages\reportlab\platypus\doctemplate.py", line 1044, in _endBuild
    if getattr(self,'_doSave',1): self.canv.save()
                                  ^^^^^^^^^^^^^^^^
      File "~\AppData\Local\pypoetry\Cache\virtualenvs\usdb-syncer-VPHy5iGw-py3.12\Lib\site-packages\reportlab\pdfgen\canvas.py", line 1301, in save
    self._doc.SaveToFile(self._filename, self)
      File "~\AppData\Local\pypoetry\Cache\virtualenvs\usdb-syncer-VPHy5iGw-py3.12\Lib\site-packages\reportlab\pdfbase\pdfdoc.py", line 212, in SaveToFile
    data = self.GetPDFData(canvas)
           ^^^^^^^^^^^^^^^^^^^^^^^
      File "~\AppData\Local\pypoetry\Cache\virtualenvs\usdb-syncer-VPHy5iGw-py3.12\Lib\site-packages\reportlab\pdfbase\pdfdoc.py", line 227, in GetPDFData
    fnt.addObjects(self)
      File "~\AppData\Local\pypoetry\Cache\virtualenvs\usdb-syncer-VPHy5iGw-py3.12\Lib\site-packages\reportlab\pdfbase\ttfonts.py", line 1336, in addObjects
    pdfFont.FontDescriptor = self.face.addSubsetObjects(doc, baseFontName, subset)
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "~\AppData\Local\pypoetry\Cache\virtualenvs\usdb-syncer-VPHy5iGw-py3.12\Lib\site-packages\reportlab\pdfbase\ttfonts.py", line 1098, in addSubsetObjects
    fontFile.content = self.makeSubset(subset)
                       ^^^^^^^^^^^^^^^^^^^^^^^
      File "~\AppData\Local\pypoetry\Cache\virtualenvs\usdb-syncer-VPHy5iGw-py3.12\Lib\site-packages\reportlab\pdfbase\ttfonts.py", line 915, in makeSubset
    glyphPos = self.glyphPos[originalGlyphIdx]
               ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
    IndexError: list index out of range
```

src/usdb_syncer/gui/report_dialog.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def _generate_report_pdf(self) -> None:
+        def on_done(_: progress.Result) -> None:
+            logger.info("PDF report created.")
+
+        fname = f"{datetime.datetime.now():%Y-%m-%d}_songlist.pdf"
+        path = os.path.join(settings.get_song_dir(), fname)
+        path = QFileDialog.getSaveFileName(self, dir=path, filter="PDF (*.pdf)")[0]
+        # dialog is hidden by main window on macOS if file picker was cancelled
+        # self.raise_()
```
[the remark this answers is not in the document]
bohning: Yes, seems to be required, unfortunately.

bohning:
> Unfortunately, it doesn't work for me. Here is the traceback after propagating the exception (see below):

Not sure what I can do about it, as it works for me on macOS.

image

bohning:
Thanks for the detailed review. I think I have addressed all issues/recommendations.

Open issues:
- Cancelling the file picker dialog closes the entire report dialog (at least on macOS).
- Not working on Windows (see traceback [here](https://github.com/bohning/usdb_syncer/pull/383#pullrequestreview-2762087974)

bohning:
> Unfortunately, it doesn't work for me. Here is the traceback after propagating the exception (see below):

I removed and re-added the font files after adding `*.ttf -text` to `.gitattributes`. This should fix the issue.

RumovZ:
Great, works now!
I have some more points.

I notice not all special characters are supported, e.g. ★. Is that a deliberate choice because the required font files would need to be far too large otherwise?

Maybe the same logic we have for artist initials in path templates should be used?

When no songs are selected Ok does seemingly nothing, if you don't pay attention to the log. Would be helpful to have a pop-up in this case or possibly just close the dialog anyway.
If a PDF was successfully created I guess most people would appreciate it opening automatically.

bohning:
> I notice not all special characters are supported, e.g. ★. Is that a deliberate choice because the required font files would need to be far too large otherwise?

I think the ★ is the only symbol that would require an additional font (Noto Sans Symbols) and additional font fallback logic, which I thought was not worth it. But the artist initial logic is simplified and harmonized with the path templates to use `unidecode` anyways.
 
> When no songs are selected Ok does seemingly nothing, if you don't pay attention to the log. Would be helpful to have a pop-up in this case or possibly just close the dialog anyway.

Good point. I added a popup with the corresponding message (but kept the dialog open).

> If a PDF was successfully created I guess most people would appreciate it opening automatically.

Also good point. The PDF now opens after successful creation (but not the JSON).

RumovZ (approved):
🚀

Issue #375, opened 2025-03-26
Command Line Interface (CLI) for scripted execution


achimmihca:
Hi, usdb_syncer is a great tool, thank you so much!

I wonder whether it would be possible to add a command line interface, such that the tool can be called from command line / terminal.

This would allow to use usdb_syncer as facade to USDB. It would enable scripting scenarios and to integrate usdb_syncer in other tools.

For example, it might look like this:
```
# show help
usdb_syncer.exe --help

# get song list index as JSON output
usdb_syncer.exe --download-index --username my_user --password my_password

# download song by id
usdb_syncer.exe --download-song 12345 --output ./my_folder/my_subfolder --username my_user --password my_password
```

Of course, the UI would not be needed for this scenario and should not start.

randompersona1:
I think it's a good idea.

Quite annoying to implement though. A lot of the syncer is based around the gui. For example, the download process currently cannot even be run from a single thread. Even worse is the settings - some functions will access settings instead of being passed the data.

Regardless, I've started some work on this.

PR #387, opened 2025-04-14
Parse `--songpath` as command line argument.

bohning:
This disables setting the songpath via environment variable (`SONG_DIR`).

### DIFF
--- a/src/usdb_syncer/gui/__init__.py
+++ b/src/usdb_syncer/gui/__init__.py
@@ -60,53 +60,59 @@ class CliArgs:
     skip_pyside: bool = False
     trace_sql: bool = False
 
+    @classmethod
+    def parse(cls) -> CliArgs:
+        parser = ArgumentParser(description="USDB Syncer")
+        parser.add_argument("--version", action="version", version=constants.VERSION)
+        parser.add_argument(
+            "--reset-settings",
+            action="store_true",
+            help="Reset all settings to default.",
+        )
 
-def main() -> None:
-    sys.excepthook = _excepthook
+        setting_overrides = parser.add_argument_group(
+            "settings", "Provide temporary overrides for settings."
+        )
+        setting_overrides.add_argument(
+            "--songpath", type=Path, help="path to the song folder"
+        )
 
-    parser = ArgumentParser(description="USDB Syncer")
-    parser.add_argument("--version", action="version", version=constants.VERSION)
-    parser.add_argument(
-        "--reset-settings", action="store_true", help="Reset all settings to default."
-    )
-
-    setting_overrides = parser.add_argument_group(
-        "settings", "Provide temporary overrides for settings."
-    )
-    setting_overrides.add_argument(
-        "--songpath", type=Path, help="path to the song folder"
-    )
-
-    if not utils.is_bundle():
         dev_options = parser.add_argument_group("development", "Development options.")
-        dev_options.add_argument(
-            "--profile", action="store_true", help="Run with profiling."
-        )
-        dev_options.add_argument(
-            "--skip-pyside", action="store_true", help="Skip PySide file generation."
-        )
         dev_options.add_argument(
             "--trace-sql", action="store_true", help="Trace SQL statements."
         )
+        dev_options.add_argument(
+            "--profile", action="store_true", help="Run with profiling."
+        )
+        if not utils.is_bundle():
+            dev_options.add_argument(
+                "--skip-pyside",
+                action="store_true",
+                help="Skip PySide file generation.",
+            )
 
-    args = parser.parse_args(namespace=CliArgs())
-    _handle_args(args)
+        return parser.parse_args(namespace=cls())
+
+    def apply(self) -> None:
+        if self.reset_settings:
+            settings.reset()
+            print("Settings reset to default.")
+        if self.songpath:
+            settings.set_song_dir(self.songpath.resolve(), temp=True)
+        if not self.skip_pyside:
+            tools.generate_pyside_files()
+        db.set_trace_sql(self.trace_sql)
+
+
+def main() -> None:
+    sys.excepthook = _excepthook
+    args = CliArgs.parse()
+    args.apply()
     utils.AppPaths.make_dirs()
     if args.profile:
         _with_profile(_run)
     else:
         _run()
-
-
-def _handle_args(args: CliArgs) -> None:
-    if args.reset_settings:
-        settings.reset()
-        print("Settings reset to default.")
-    if args.songpath:
-        settings.set_song_dir(args.songpath.resolve(), temp=True)
-    if not args.skip_pyside:
-        tools.generate_pyside_files()
-    db.set_trace_sql(args.trace_sql)
 
 
 def _run() -> None:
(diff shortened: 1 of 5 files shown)


src/usdb_syncer/gui/__init__.py
```diff
@@ -103,17 +108,26 @@ def _with_profile(func: Callable[[], None]) -> None:
...
+    song_dir = settings.get_song_dir()
+    if songpath := args.songpath:
+        if Path(songpath).exists():
+            song_dir = Path(songpath).resolve()
+        else:
+            logging.warning(
+                f"Song directory '{songpath}' does not exist, using "
+                f"{song_dir} instead."
+            )
+    mw.lineEdit_song_dir.setText(str(song_dir))
```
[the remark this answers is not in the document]
bohning: Ah, got it. Like this?

src/usdb_syncer/gui/mw.py
```diff
@@ -149,8 +150,18 @@ def _setup_toolbar(self) -> None:
     def _setup_shortcuts(self) -> None:
         gui_utils.set_shortcut("Ctrl+.", self, lambda: DebugConsole(self).show())
 
-    def _setup_song_dir(self) -> None:
-        self.lineEdit_song_dir.setText(str(settings.get_song_dir()))
+    def _setup_song_dir(self, songpath: str | None) -> None:
+        self.song_dir = settings.get_song_dir()
```
RumovZ, line 154: song_dir is not kept in sync with lineEdit_song_dir. And more importantly: --song-path is not used for downloading.
It would be much easier if there was a single source of truth. What if --songpath also updates the setting? I understand that's less versatile, but also much less error-prone.

bohning: Not sure if that's not confusing. I would expect that passing `--songpath`*temporarily` overwrites the song_dir, and if I restart the syncer with argument, it will start with the song_dir stored in the settings.

RumovZ: I don't really mind either way, as long as it's consistent and the code is clear.

randompersona1:
I made some additions in agreement with @bohning . They should provide the functionality we want - temporarily overriding the settings for a single session only. 

Typing gets really messy, so I'm not very confident that this works perfectly right now. We should definitely have some unit tests to be sure.

On that note, it might be worth considering writing our own settings backend instead of using QSettings. A simple json file would provide more consistency across platforms and types.

src/usdb_syncer/gui/__init__.py
```diff
@@ -55,6 +62,27 @@ def main() -> None:
...
+                    songpath = Path(value).resolve(strict=True)
+                    if not songpath.is_dir() or not songpath.exists():
+                        raise ValueError(f"Path is invalid: {value}")
+                    settings.set_song_dir(songpath, temp=True)
+                except (OSError, RuntimeError) as e:
+                    raise ValueError(
+                        f"Invalid songpath provided: {value}. Error: {e}"
+                    ) from e
+            case _:
+                assert False, f"Unhandled argument: {arg}"
```
RumovZ, line 83: argparse can handle a lot of this if you pass a dataclass to be used instead of the generic Namespace.

src/usdb_syncer/settings.py
```diff
@@ -64,6 +65,93 @@ def ffmpeg_is_available() -> bool:
...
+        return key in cls._temporarySettings
+
+
+class _Settings:
+    """Settings for the USDB Syncer.
+
+    This class is a singleton that provides access to the settings of the USDB Syncer.
+    """
+
+    _lock = threading.RLock()
```
RumovZ, line 104: Why the lock?

RumovZ (commented):
Looks good with _TemporarySettings!

What exactly do you mean by consistency? QSettings has some edge cases regarding primitive types, but as far as I'm aware we have them all covered.
A JSON solution would also need to implement things like buffering and marshalling to be on par with QSettings. To justify the cost and risk of bugs, there would have to be some decisive advantages.

src/usdb_syncer/settings.py
randompersona1: I thought it more readable, and considering settings shouldn't be written often, much less simultaneously, the performance overhead is minimal to nonexistent.

It could be removed if we switched some methods around to ensure there are no race conditions.

RumovZ: I guess a race condition could theoretically appear in code like `if _TemporarySettings.contains(key): _TemporarySettings.remove(key)`, but that could be avoided using atomic operations like `get` and `pop`.
Since we rarely if ever access settings from a worker thread, I don't think we have to worry much about it.

randompersona1: I think it's worth the tiny bit of effort to make the settings thread-safe in general. I'm not certain, but I think at the very least `get_song_folder()` is called from different threads.

I realize this is getting a bit far-fetched, but the google style guide [explicitly forbids relying on the atomicity of dicts](https://google.github.io/styleguide/pyguide.html#218-threading). It's a CPython feature, not a language feature, and reliant on the GIL, which is supposed to be removed at some point soon<sup>TM</sup>. I know it'd be trivial to change it if it ever mattered, but I'd rather do it "correctly" now.

I don't mind being overridden though if you truly don't think it's neccessary. In that case, I'll still make some changes so that we only rely on the dict and QSettings to be atomic. That won't guarantee that the order is correct, but it means we should be safe from any exceptions or dropped data.

randompersona1:
You're right about QSettings. I just really hate that block of code that returns the correct type of setting.

Regarding platform compatibility, I also didn't know at the time that QSettings allowed setting the file format manually. I don't like storing settings in the windows registry because they're so hard to delete manually, but this has given me some other ideas. Let's leave it at that for now.

src/usdb_syncer/settings.py
RumovZ: I see! Wasn't aware it's only a CPython thing. 😮
As another idea, we could also make settings thread-local to enforce them being passed from the main thread. But I'm just spitballing, perfectly fine with the lock for now.

src/usdb_syncer/gui/__init__.py
randompersona1: I don't know. I don't really like the style either way tbh. A dataclass would still be untyped, which is annoying, and would introduce a third command reference. So we'd have to update three things if we wanted to change a name.

I'm open for suggestions, but in the meantime, I'll research how other cli tools do this nicely. Already had a look at yt-dlp and it's horrific.

RumovZ:
```py
from argparse import *
from pathlib import *

class Args:
    songdir: Path | None = None

parser = ArgumentParser()
parser.add_argument("--songdir", type=Path)
args = Args()
parser.parse_args(["--songdir", "/foo/bar"], namespace=args)
```

This gives you a typed object without any literal string matching like above. Seems flawless to me.

randompersona1: Ugh, I'm an idiot. I was set on continuing to iterate through the dataclass like right now in _handle_args, which gets rid of the types again.

randompersona1:
I'm happy with this changes now. I took the liberty of adding a couple of other simple command line arguments as well. If we want to implement #375, we'd have to make some flow changes so we can control whether the gui launches while keeping other initialization steps, but I think that's out of the scope for this PR.

bohning:
Thanks a lot for picking up, improving on and extending my initial changes!

RumovZ:
> I took the liberty of adding a couple of other simple command line arguments as well.

What do you think about converting the remaining two env vars into command line arguments as well then?

RumovZ (approved):

PR #394, opened 2025-04-23
Use ruff as main linter and formatter

randompersona1:
This PR tracks adding ruff as the main linter and formatter.

So far, I've only changed the configuration files. I want to fix the linting errors in one go (since there are a lot), so I want to be sure that we're on the same page regarding the linting rules. They're configured in `tool.ruff.lint` in the pyproject.toml, and all available rules are at https://docs.astral.sh/ruff/rules/.

The current configuration enforces a much more strict code style than before. I do think this is sensible though.

The ruff linter is launched through the normal tox pipeline or with `poetry run ruff check .`
The ruff formatter (replacing black) is launched with `poetry run ruff format .` and accepts `--check` or `--diff` analogue to black.

To skip a check in a line or module, ruff uses the following comment scheme: `# noqa: E101`, where `E101` can be any ruff rule.

Ruff also has an official vscode extension so linting errors show up directly in the IDE.


RumovZ (commented):
It's a pity you can't ignore rules by name. That at least partly defeats the purpose of demanding an explicit ignore in the first place.

There are a few interesting new findings, but most are false positives we should silence. How do you want to proceed? Do you want to discuss specific rules to ignore?

pyproject.toml
```diff
@@ -99,18 +92,34 @@ disable = [
...
+    ".mypy_cache",
+    ".pytest_cache",
+    ".ruff_cache",
+    ".tox",
+    ".venv",
+    "build",
+    "dist",
+    "venv",
+    "forms",
+    "resources",
```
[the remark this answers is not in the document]
randompersona1: Ah, that's perfect! I was looking for an option like this, but I couldn't find it.

randompersona1:
The main thing to discuss for me is whether we *don't* want to implement some of the rules I added, or if we want to make any other wide-reaching configuration changes that would specifically affect linting.

If not, I'll start fixing the linting errors by hand. We can take some shortcuts, like I did with `assert` statements in tests only, but as you said, many are genuine style improvements, and most others are a matter of suppressing the error.

I suppose it's also worth considering adding additional rules straight away in this PR if there are obvious improvements I missed. The last commit added `COM, G, LOG, Q`, but it's honestly hard to keep track of which ruleset does what and which are already included, so there are surely other sensible ones.

RumovZ:
Here are my opinions to all the rules we currently violate.

Helpful; code should be fixed.
- A001 Variable `id` is shadowing a Python builtin
- B007 Loop control variable `idx` not used within loop body
- B008 Do not perform function call `MetaTags` in argument defaults; instead, perform the call within the function, or read the default from a module-level singleton variable
- B905 [*] `zip()` without an explicit `strict=` parameter
- C400 Unnecessary generator (rewrite as a list comprehension)
- C401 Unnecessary generator (rewrite as a set comprehension)
- C408 Unnecessary `tuple()` call (rewrite as a literal)
- C901 `__str__` is too complex (13 > 10)
- E226 [*] Missing whitespace around arithmetic operator
- E266 [*] Too many leading `#` before block comment
    - A bit overeager, but fine ...
- E501 Line too long (100 > 88)
- E711 Comparison to `None` should be `cond is None`
- E713 [*] Test for membership should be `not in`
- E741 Ambiguous variable name: `l`
- LOG002 Use `__name__` with `logging.getLogger()`
- LOG015 `error()` call on root logger
- N815 Variable `_temporarySettings` in class scope should not be mixedCase
- PTH109 `os.getcwd()` should be replaced by `Path.cwd()`
- PTH118 `os.path.join()` should be replaced by `Path` with `/` operator
- PTH120 `os.path.dirname()` should be replaced by `Path.parent`
- PTH122 `os.path.splitext()` should be replaced by `Path.suffix`, `Path.stem`, and `Path.parent`
- PTH123 `open()` should be replaced by `Path.open()`
- PTH204 `os.path.getmtime` should be replaced by `Path.stat().st_mtime`
- PTH207 Replace `glob` with `Path.glob` or `Path.rglob`
- RUF005 Consider `(*changed, self)` instead of concatenation
- RUF010 [*] Use explicit conversion flag
- RUF012 Mutable class attributes should be annotated with `typing.ClassVar`
- RUF046 [*] Value being cast to `int` is already an integer
- S108 Probable insecure usage of temporary file or directory: "/tmp/\_MEI"
- S108 Probable insecure usage of temporary file or directory: "/tmp/\_MEI12345:/usr/lib:/usr/local/lib"
- RUF018 Avoid assignment expressions in `assert` statements
- TRY003 Avoid specifying long messages outside the exception class

Helpful, but needs to be silenced (like with pylint before).
- S102 Use of `exec` detected
- N802 Function name `closeEvent` should be lowercase
    - I feel like pylint was a little bit better at detecting if the violating name is an overwrite (but also suddenly stopped recognising overwrites sometimes). Still, I think we should keep the rule and ignore overwrites individually.
- S311 Standard pseudo-random generators are not suitable for cryptographic purposes
- TRY400 Use `logging.exception` instead of `logging.error`
    - We log the traceback to a different level, so this doesn't make sense for us. But maybe we should write a helper for this and only silence ruff there, so we still get notified when there was an actual oversight.

Same as above, but let's silence it on the module level.
- RUF001 String contains ambiguous ` ` (NARROW NO-BREAK SPACE). Did you mean ` ` (SPACE)?
- RUF003 Comment contains ambiguous `‘` (LEFT SINGLE QUOTATION MARK). Did you mean ``` (GRAVE ACCENT)?
- W291 Trailing whitespace

Let's exclude these:
- G003 Logging statement uses `+`
    - I don't see any reason why this would be bad in our use case, and it's more readable.
- G004 Logging statement uses f-string
    - Same as G003.
- S101 Use of `assert` detected
    - True, `assert` shouldn't be used for validation, but it's great for cases where you know more than the type checker. The alternatives would be:
        1. Check manually and raise a custom exception. This is just too bothersome to handle a case you know (or strongly believe) cannot happen.
        2. Use `cast` or `type: ignore`. This is objectively worse, because `assert` helps at least if optimizations are disabled (which is basically always for our use case).
- S404 `subprocess` module is possibly insecure
    - Well, we wouldn't import `subprocess` if we didn't intend to use it.
- S603 `subprocess` call: check for execution of untrusted input
    - This is pointless because it seems to flag just everyting.
- S606 Starting a process without a shell
    - Same as S404.
- S607 Starting a process with a partial executable path
    - Not entirely sure, but I think we simply can't know the absolute path in many cases.
- S608 Possible SQL injection vector through string-based query construction
    - Same as S603.

<hr>

All in all, I'm surprised in a good away about the proportion of justified versus spurious findings.

Some of these pylint definitely shoud have picked up, too. I wonder why it didn't and if it's also true the other way around (stuff pylint finds, but ruff doesn't, despite both having rules for it).

Regarding newly added rule sets, let's pin ruff so we can enable new rules periodically and deliberately. It's annoying if you just want to update some dependency and are hit with a bunch of new lints.

RumovZ:
> We should look at at asserts specifically in code reviews if we're taking it out.

Agreed. In that it is like `cast()`, `# noqa` etc.

>  we could also activate ruff's pylint rules

Maybe we can try this in a new PR, or after the ones above are fixed.

If you want help, I could start working on the code fixes (first list in my previous comment).

RumovZ (approved):

RumovZ:
Looks good from my side! Do you want to add anything else? I've looked at the pylint rules and they don't seem to provide any additional value.

bohning (approved):
Great stuff!

src/usdb_syncer/utils.py
```diff
@@ -230,7 +231,7 @@ def next_unique_directory(path: Path) -> Path:
 
 
 def is_name_maybe_with_suffix(text: str, name: str) -> bool:
-    """True if `text` is 'name' or 'name (n)' for the provided `name` and some number n."""
+    "True if `text` is 'name' or 'name (n)' for the provided `name` and some number n."
```
randompersona1, line 234: I'm impressed, but from https://peps.python.org/pep-0257/:

> For consistency, always use """triple double quotes""" around docstrings. 

I'm surprised this isn't enforced by either the linter or the formatter.

randompersona1 (commented):
Looks very nice! I only have minor questions

src/usdb_syncer/utils.py
RumovZ: This is one of the suppressed rules ([Q002](https://docs.astral.sh/ruff/rules/bad-quotes-docstring/#bad-quotes-docstring-q002)). No idea why the formatter doesn't enforce triple quotes, though.
I first tried to fix this by moving the closing quotes to the next line how you would normally do. But the formatter kept shifting them back.
We could add a redundant line break within the text, but I would just leave it for now. At some point, the formatter will probably be fixed and this resolves itself.

RumovZ:
Thanks, @randompersona1!

PR #407, opened 2025-05-06
Fix db expression tree

### LINKED ISSUE #406: Crash on Windows
On Windows the app is crash!
Version: 0.12.1 and v0.13.0 the same crash

Image

`sqlite3.OperationalError: Expression tree is to large`

Please do me a favor and test the program if anyone gets the idea to download all the songs.

randompersona1:
Fixes #406. I also added some tests to ensure that this error is caught in the future.

Finally, I added an option for pytest to mark tests as slow, which requires an additional cli argument. It isn't run by default locally because test failures should be rare, but the GH Actions pipeline, does also test those.

### DIFF
--- a/src/usdb_syncer/db/__init__.py
+++ b/src/usdb_syncer/db/__init__.py
@@ -820,7 +820,10 @@ class ResourceFileParams:
 
 
 def delete_resource_files(ids: Iterable[tuple[SyncMetaId, ResourceFileKind]]) -> None:
-    for batch in batched(ids, _SQL_VARIABLES_LIMIT // 2):
+    for batch in batched(ids, 980):  # 998 is the maximum batch size we can use. Since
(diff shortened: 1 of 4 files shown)


src/usdb_syncer/db/__init__.py
```diff
@@ -820,7 +820,10 @@ class ResourceFileParams:
 
 
 def delete_resource_files(ids: Iterable[tuple[SyncMetaId, ResourceFileKind]]) -> None:
-    for batch in batched(ids, _SQL_VARIABLES_LIMIT // 2):
+    for batch in batched(ids, 980):  # 998 is the maximum batch size we can use. Since
```
bohning, line 823: Maybe this should be a defined constant (along with the explanatory comment).

bohning (approved):

PR #408, opened 2025-05-06
Remove unused dependancies

randompersona1:
I also updated `pyinstaller-hooks-contrib`, `cryptography`, and `virtualenv`


bohning (approved):

pyproject.toml
```diff
@@ -63,6 +60,7 @@ pyinstaller-hooks-contrib = "*"
 altgraph = "*"
 macholib = {version = "*", markers="sys_platform == 'darwin'"}
 pywin32-ctypes = {version = "*", markers="sys_platform == 'win32'"}
+setuptools = { version = "^75.8.0", optional = true }
```
RumovZ, line 63: What do we need it in bundling for?

RumovZ: It was actually only recently added by @bohning in 96cf1fde15900ef4db2fd0a18af4c5ac9dc40ccc. Not sure why. I guess we can just drop it?

bohning: Can't remember why I added that. I guess we can drop it.

randompersona1: I removed it and updated to latest version (80.3.1).

PR #415, opened 2025-05-10
Add user/connection status info to statusbar.

bohning:
Open issues:
- get actual logged in user name
- get actual connection status
- signal/slot setup for changes in connection status


RumovZ:
I've added `icons.py` for centralized and theme-aware icons. You can then use descriptive filenames and semantic names in Python.

Since every user will only ever see one icon and the association between icon and role is arbitrary anyway, I think there's no point in using different role icons.

In general, I find this is a great idea to reduce reliance on information scattered across the log. Some more ideas to expand on this:
- Status should have tooltips explaining the details, and be clickable to get to the relevant settings.
- Also include status from YouTube, Fmpeg etc. so the user can see at a glance if they're ready to go.

randompersona1:
I've implemented some of the functionality, but I find `usdb_scraper` quite inconvenient for this purpose. It also lacks a a lot of error handling. Any ideas on how to refactor it nicely? Perhaps a single class that has to be instantiated for every session?

PR #416, opened 2025-05-11
Use bash when retrieving version.

bohning:


randompersona1 (approved):
This really shouldn't make a difference, but I don't mind either way

PR #417, opened 2025-05-14
Fix extraction of allowed countries.

bohning:
Switching from `html.parser` to `lxml` for BeautifulSoup fixes handling of malformed HTML (`<tr>` and `<td>` elements are never closed).


RumovZ (approved):

tests/unit/test_get_allowed_countries.py
```diff
@@ -0,0 +1,95 @@
...
+    return (
+        len(country_code) == 2
+        and country_code.isalpha()
+        and country_code in country_codes
+    )
+
+
+def get_test_resources() -> list[tuple[Path, Path]]:
+    """Get pairs of HTML and JSON files for testing."""
+    resources_dir = Path(__file__).parent.parent / "resources" / "html" / "polsy-org-uk"
```
randompersona1, line 31: I can't think of a way to use the `resource_dir` fixture, but if there is one, I'd be glad to change.

tests/unit/test_get_allowed_countries.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+@pytest.mark.parametrize("html_file, json_file", get_test_resources())
+def test_get_allowed_countries(
+    html_file: Path, json_file: Path, resource_dir: Path, country_codes: list
+) -> None:
+    """Test get_allowed_countries with different HTML/JSON pairs."""
+    # Load test data
+    with html_file.open("r", encoding="utf-8") as f:
+        html_content = f.read()
```
RumovZ, line 59:
```suggestion
    html_content = html_file.read_text(encoding="utf-8")
```

tests/unit/test_get_allowed_countries.py
```diff
@@ -0,0 +1,95 @@
...
+    with json_file.open("r", encoding="utf-8") as f:
+        expected_data = json.load(f)
+
+    # Extract expected allowed countries
+    expected_allowed_countries = None
+    if expected_data.get("none") is False and "allowed" in expected_data:
+        expected_allowed_countries = expected_data["allowed"]
+
+    # Mock the requests.get call
+    with patch("requests.get") as mock_get:
```
RumovZ, line 70: `get_allowed_countries` does both fetching and extracting. If this was cleaned up you wouldn't need to mock here.

tests/unit/test_get_allowed_countries.py
RumovZ: There's also the option to list and import resources explicitly as Python variables as I've recently done for `gui.resources`. But as it currently stands, I have no issues with `get_test_resources()`.

tests/unit/test_get_allowed_countries.py
randompersona1: I don't quite see the point of simply using two methods. I think readability suffers more that way.

However, I could see refactoring `usdb_scraper` to create a unified interface for making web requests, automatically setting an appropriate user-agent and cookies, and handling retries, logging, errors, etc.. Feels a bit outside of the scope of this PR though.

RumovZ: The point is [separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns) in general, and here specifically, the ability to test parsing without having to mock fetching.

tests/unit/test_get_allowed_countries.py
RumovZ: Yeah, I meant in the future we can potentially drop that fixture and use type safe imports instead like `from ..resources.html import ...`. But until then I don't think duplicating the path here is so bad.

PR #420, opened 2025-05-21
Concentrate and improve networking

randompersona1:
In reference to my comment [here](https://github.com/bohning/usdb_syncer/pull/415#issuecomment-2871683317), and in general too, I reworked much of the actual networking aspects of `usdb_scraper.py`. The new interface(s) in `net.py` allow for unifying error handling and retries.

I intend to rewire all other networking calls to this new module, so that the only place `requests` is imported is here.

This is unfinished, but I'm opening the PR for feedback. Remaining work includes:
- [ ] Using `net.py` for all other requests
- [ ] Sending events in specific instances (#415)
- [ ] Add (or sometimes re-add) debug logging
- [ ] Add error handling when a request fails
- [ ] (potentially) add (configurable?) retries for when a request fails
- [ ] Maybe remove `logged_in` attribute of `UsdbSession` and just use username

### DIFF
--- a/src/usdb_syncer/usdb_scraper.py
+++ b/src/usdb_syncer/usdb_scraper.py
@@ -1,18 +1,14 @@
 """Functionality related to the usdb.animux.de web page."""
 
 import re
-import time
 from collections.abc import Iterator
 from datetime import datetime
-from enum import Enum
-from typing import Any, assert_never
+from typing import Any
 
 import attrs
-import requests
 from bs4 import BeautifulSoup, NavigableString, Tag
-from requests import Session
 
-from usdb_syncer import SongId, errors, settings
+from usdb_syncer import SongId, errors
 from usdb_syncer.constants import (
     SUPPORTED_VIDEO_SOURCES_REGEX,
     Usdb,
@@ -21,10 +17,13 @@ from usdb_syncer.constants import (
     UsdbStringsFrench,
     UsdbStringsGerman,
 )
-from usdb_syncer.logger import Logger, logger, song_logger
+from usdb_syncer.logger import Logger, song_logger
 from usdb_syncer.usdb_song import UsdbSong
-from usdb_syncer.utils import extract_youtube_id, normalize
+from usdb_syncer.utils import extract_youtube_id
 
+WELCOME_REGEX = re.compile(
+    r"<td class='row3' colspan='2'>\s*<span class='gen'>([^<]+) <b>([^<]+)</b>"
+)
 SONG_LIST_ROW_REGEX = re.compile(
     r'<td(?:.*?<source src="(?P<sample_url>.*?)".*?)?></td>'
     r'<td onclick="show_detail\((?P<song_id>\d+)\)".*?>'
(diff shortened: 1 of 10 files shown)


randompersona1:
From my side, this is finished. The main improvement outside `usdb_scraper` is that all requests using `net._GenericSession` now have automatic error handling. Custom handling on a site-to-site basis should be achieved by subclassing `net.SyncerSession` (as seen with `usdb_scraper.UsdbSession`.

src/usdb_syncer/utils.py
```diff
@@ -263,7 +254,10 @@ def add_to_system_path(path: str) -> None:
 
 def normalize(text: str) -> str:
     """Return the Unicode NFC form of the string."""
-    return unicodedata.normalize("NFC", text)
+    t = time.time()
```
RumovZ, line 257: Committed by mistake?

src/usdb_syncer/usdb_scraper.py
```diff
@@ -40,95 +44,235 @@
...
+        )
+        if UsdbStrings.NOT_LOGGED_IN not in text:
+            if self.username != username:
+                self.username = username
             return True
-        logger.error(f"Login to USDB with user '{auth[0]}' failed!")
-    else:
-        logger.warning(
-            "Not logged in to USDB. Please go to 'Synchronize > USDB Login', then "
-            "select the browser you are logged in with and/or enter your credentials."
```
RumovZ, line 61: While the menu path is outdated, I find this warning still valuable.

src/usdb_syncer/usdb_scraper.py
```diff
@@ -40,95 +44,235 @@
...
+    def cookie_login_exists(self) -> bool:
+        """Check if a login exists using session cookies.
+
+        Uses the cookies stored in this session.
+
+        Returns:
+            True if login was successful, False otherwise.
+
+        """
+        username = cast(dict[str, str], json.loads(self.get("whoami.php"))).get(
```
RumovZ, line 127: If the request fails, we get a misleading JSON error. This also has the implication that the syncer won't launch without a network connection anymore.

src/usdb_syncer/usdb_scraper.py
```diff
@@ -40,95 +44,235 @@
...
+                f"Request failed with status code {response.status_code} for URL: "
+                f"{response.url}"
+            )
+            return ""
+        return normalize(response.text)
+
+    @override
+    def conn_failed(self, url: str, error: RequestException) -> str:
+        """Handle connection failure."""
+        self.connected = False
```
RumovZ, line 73: Swallowing all these exceptions that were raised before will have some unforeseen consequences (below is just the one I noticed immediately).

src/usdb_syncer/usdb_scraper.py
```diff
@@ -40,95 +44,235 @@
...
+        """
+        username = cast(dict[str, str], json.loads(self.get("whoami.php"))).get(
+            "username", ""
         )
-    return False
+        if username and self.username != username:
+            self.username = username
+        return bool(username)
 
+    def establish_login(self, auth: Tuple[str, str] | None = None) -> bool:
```
RumovZ, line 134:
```suggestion
    def establish_login(self, auth: tuple[str, str] | None = None) -> bool:
```

src/usdb_syncer/usdb_scraper.py
randompersona1: Ah. I expected json.loads("") to create an empty json object. Should've checked that.

src/usdb_syncer/usdb_scraper.py
randompersona1: True, I forgot to handle this. I suppose there are two solutions:

1. Raise the exception here
2. Add an `if text := self.get` for every get request and deal with the error explicitly along the call stack. Probably involves changing `song_loader` pretty extensively.

Any other ideas? I'm not too hung up on this method in general, but I'm also not a fan of the previous solution of just erroring. The pain of web requests in general, I guess.

RumovZ: The syncer is built to deal with most unexpected exception. To handle things more gracefully, I don't see an alternative to catching exceptions on a case-by-case basis.
On a related note, I don't think the session class is the ideal level to define response handling and type. They could easily differ between requests.

PR #421, opened 2025-05-21
Instrumental version

bohning:
This adds the option to separate the audio file into two stems (instrumental, vocals) using [demucs](https://pypi.org/project/demucs/). The stems are converted to the user-selected target format/codec and are referenced in the header of the txt file with the appropriate tags (#INSTRUMENTAL, #VOCALS). This requires setting the format version to at least 1.1.0 (earlier versions do not include those headers).
The upcoming USDX release already supports switching between the regular audio track (#MP3/#AUDIO) and the instrumental track (#INSTRUMENTAL) by pressing `K` during gameplay.
Unfortunately, this does not make use of GPU acceleration, thus may take quite bit depending on CPU processing power.


Baklap4:
@bohning thanks for this feature!

I'm just wondering will the instrumental songversions in usdb be eventually cleaned up so there's no duplicate song  for the instrumental and normal one?

+ A question for here; would it be possible to specify a youtube (or any other) source so we don't need to demucs everything?

src/usdb_syncer/settings.py
```diff
@@ -437,6 +438,25 @@ def __str__(self) -> str:
...
+    HTDEMUCS_FT = "htdemucs_ft"
+
+    def __str__(self) -> str:
+        match self:
+            case AudioStemSeparation.DISABLE:
+                return "disabled"
+            case AudioStemSeparation.HTDEMUCS:
+                return "htdemucs (faster, lower quality)"
+            case AudioStemSeparation.HTDEMUCS_FT:
+                return "htdemucs_ft (slower, higher quality)"
```
randompersona1, line 455: While most people will probably just try out both, I feel it's worth mentioning that the speed difference is massive (4x), while the quality difference is fairly small. Maybe just add `much` in front of `faster` and `slower`?

pyproject.toml
```diff
@@ -70,7 +74,7 @@ pywin32-ctypes = {version = "*", markers="sys_platform == 'win32'"}
 usdb_syncer = "usdb_syncer.gui:main"
 
 [tool.poetry-dynamic-versioning]
-enable = true
+enable = false
```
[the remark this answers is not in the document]
bohning: Same here, don’t know how that was changed.

src/usdb_syncer/settings.py
bohning: Thanks for the review and feedback. I think I fixed everything now.

pyproject.toml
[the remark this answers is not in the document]
RumovZ: It's changed by dynamic versioning, which changes it back as well. Maybe you've interrupted it or checked it in while the `poetry` command was still running.

pyproject.toml
```diff
@@ -23,21 +23,25 @@ python = ">3.10,<3.13"
...
 pillow = ">=10"
 PySide6 = "*"
+python-ffmpeg = "^2.0.12"
 reportlab = "*"
 requests = "*"
 rookiepy = "^0.5.3"
 send2trash = "*"
 sounddevice = "^0.5.1"
 soundfile = "^0.13.1"
+torch = "2.2.2"  # 2.2.2 is latest version available for macosx_x86_64 (Intel Macs)
```
RumovZ, line 43: Doesn't it automatically resolve to a compatible version depending on the platform?

src/usdb_syncer/song_loader.py
```diff
@@ -668,6 +711,57 @@ def _maybe_write_audio_tags(ctx: _Context) -> None:
...
+    ctx.out.instrumental.new_fname = (
+        f"{ctx.locations.filename()} [INSTR].{audio_options.format.value}"
+    )
+
+    vocals_input = str(ctx.locations.stem_separation_vocals_path(model))
+    vocals_output = f"{ctx.locations.temp_path()} [VOC].{audio_options.format.value}"
+    transcode_audio(audio_options, vocals_input, vocals_output)
+    ctx.out.vocals.resource = ctx.out.audio.resource
+    ctx.out.vocals.new_fname = (
+        f"{ctx.locations.filename()} [VOC].{audio_options.format.value}"
```
RumovZ, line 748: Isn't this just `vocals_output.name`?

pyproject.toml
bohning: Unfortunately for some reason, it doesn't. We are [not alone](https://github.com/python-poetry/poetry/issues/7597) with this issue, it seems.

RumovZ: Does only pinning the version for macOS 10 work?

src/usdb_syncer/song_loader.py
bohning: You mean `Path(vocals_output).name`, right?

pyproject.toml
RumovZ: Hm, if it's that hardware-dependent it's also probably a nightmare to bundle. This should be solved before the merge in order not to block the next release.

src/usdb_syncer/song_loader.py
RumovZ: Ah, yes. But I would build the correct path object first: `vocals_output = ctx.locations.temp_path(ext=f" [VOC].{audio_options.format.value}")`.

pyproject.toml
randompersona1: Oh yes, lots of fun.

This issue might be bigger than a review comment chain, but oh well.
To elaborate on @bohning's answer, and to sort my own thoughts, the chain of issues is as follows:
1. We need as recent of a release of pytorch as we can get, because they have very limited python version compatibility: each python version requires a different pytorch build.
2. The last pytorch release for intel macs is `2.2.2` (https://github.com/pytorch/pytorch/issues/114602). If we really wanted to, we might be able to build later versions ourselves, but that sounds horrific. I haven't found any existing builds I deemed trustworthy.
3. `2.2.2` is a convenient version, because it is the first minor version to support python3.12. So we couldn't use anything `<2.2.0` anyways. However, if we wanted to upgrade to any python version `>3.12`, we'd need to drop support for x86 mac and bump pytorch to a more recent release. 
4. **(this is where currently unfixed problems start)**: The pypi windows builds are built with cuda dependencies. As they are very large, and do not apply to everybody, we should not bundle with cuda, or at least supply a bundle without cuda. Despite trying for a while, @bohning and I could not get poetry to grab the macos cpu-only files from pypi, while using the windows cpu-only files from pytorch's download page. We tested with `uv` eventually (https://github.com/randompersona1/demucs-testing), and that worked. Seems to be an issue with poetry (https://github.com/python-poetry/poetry/issues/7597) not parsing certain markers correctly. The only solution I currently see here is actually switching to `uv`
5. Despite what the demucs documentation says, running with cuda in in torch versions `>2.1.2` is possible, but requires  a package called `soundfile` (which we use already). I do not know what the significance of cuda versions is, but `torch=2.2.2` is built around a fairly old one (cu121, or toolchain version 12.1).
6. We need to figure out a strategy for bundling/packaging with/without cuda. Personally, I'd be fine with having it as a poetry/uv extra and force people who want cuda to run from source or bundle themselves.

bohning: Even though I am personally affected by this, I have no issue with dropping support for x86 macs. It is painfully slow to separate stems on old machines like this, so I wouldn't use the feature anyways except for maybe a handful carefully selected songs.

randompersona1: I think it makes sense to offer non-instrumental versions of the syncer, as torch, even in CPU-only, is at least 100MB. 

The way to achieve this is probably by making demucs and torch optional dependencies (we would just not allow x86 mac to install those). Of course, we'd need to handle demucs not being available in-code - I'd just grey out the `Instrumental` option. We would then supply a single non-demucs bundle for x86 mac, and two bundles for every other os. 

It is then also possible to create more dependency extras for hardware acceleration (https://python-poetry.org/docs/dependency-specification/#extra-environment-marker). I don't think we should bundle those, but making them available via extras would be awesome for pypi users.

RumovZ: Thanks for the write-up, @randompersona1. @bohning, that wouldn't solve bundling on the other platforms, though.
I agree that excluding this feature from the bundle seems reasonable. Users still wouldn't need to run from source. Installing from PyPI should also work.

RumovZ: (Hadn't read @randompersona1's second comment.)
Is it really worth it this amount of complexity just to use this one feature from the bundle, too? Makes it also more difficult for the user to find the bundle they need. And it doesn't scale: What if there is another optional feature in the future? Would we then have 4 bundles per platform?

randompersona1: I mean, at that point, it's barely any extra complexity, just another matrix step with different poetry arguments. But you're right in that it doesn't scale.

So for example for the windows version, would you bundle with torch, increasing the size of the binary to >200MB, or tell people to use pypi?

RumovZ: I would tell them to use PyPI so the majority of users don't have to download a bunch of stuff they don't need, and we don't make bundling even more fragile than it already is.

randompersona1:
I implemented the dependency specification as we discussed. The default install doesn't have demucs at all, which is handled at runtime. Currently missing is some kind of information in the gui for this.

To add demucs, add `--extras instrumental-cpu`/`--extras instrumental-cuda` to `poetry sync` or `poetry install`. I am 95% sure that this will allow pypi users to specifiy `pipx install usdb_syncer[instrumental-cuda]`, but I haven't tested this yet.

On that note, poetry seems extremely unreliable for me here. First of all, for some reason, it still finds and selects torch `2.2.2`, even though `2.7.0` is set explicitly everywhere. Furthermore, while both cases work from a clean install, switching from `instrumental-cpu` to `instrumental-cuda` (and in reverse) can fail in various ways. Sometimes, it selects said version `2.2.2`. Sometimes, it doesn't switch at all. Sometimes, it will report that torch could not be installed. I have no idea why any of these happen or how to fix them. I _think_ the `pyproject.toml` is set up correctly, but that only leaves a bug in the poetry resolver. Unsure how to proceed.

@RumovZ , can you try to reproduce the errors? Perhaps something is wrong with my setup or configuration.
@bohning , could you test on your mac? I don't know what will happen when you try to use `instrumental-cpu` or `instrumental-cuda`. I want it to return "Could not install because torch2.7.0 is not available for your machine", but I suspect it will install 2.2.2 or something and that's why it interferes with everything else.

PR #422, opened 2025-05-23
Fix incorrect error message

DeinAlptraum:
There is no "Synchronize" item in the menu bar. The "USDB Login" entry can be found under "USDB".


RumovZ:
Thanks, it's already fixed in #420. I'll close this to avoid merge conflicts.

PR #425, opened 2025-05-26
Do not delete #VIDEOGAP, but keep warning.

bohning:
There are a few exceptions where the video does not align to its own audio track, so a #VIDEOGAP fixes async audio and video of the same source. Therefore, we keep the #VIDEOGAP even if audio and video stem from the same source, but also keep the warning, because it the described situation is rare and the usage of #VIDEOGAP is most likely not intended or the resources are incomplete.

### DIFF
--- a/src/usdb_syncer/song_txt/headers.py
+++ b/src/usdb_syncer/song_txt/headers.py
@@ -191,12 +191,12 @@ class Headers:
                 or meta_tags.audio == meta_tags.video
             ):
                 logger.warning(
-                    "This song contains a non-zero #VIDEOGAP, which only makes sense "
-                    "if different resources for audio and video are specified, which "
-                    "is not the case here. This should be fixed in USDB. Removing "
-                    "#VIDEOGAP in local text file."
+                    "This song contains a non-zero #VIDEOGAP, which usually aligns "
+                    "audio and video from different sources. For this song, audio and "
+                    "video come from the same source, so the #VIDEOGAP is most likely "
+                    "incorrect unless the video track is indeed not in sync with its "
+                    "own audio track. Please check this manually."
                 )
-                self.videogap = None
 
 
 def _set_header_value(kwargs: dict[str, Any], header: str, value: str) -> None:


randompersona1 (approved):
Do have a way of knowing if this will break any existing songs?

bohning:
Not sure if I understand your question. This feature of locally deleting the #VIDEOGAP if source of audio and video is identical was only recently introduced. I thought it would *never* be necessary to align a video to its own audio track and thus introduced the feature, because a #VIDEOGAP (that was *meant* for audio and video from different sources) actually *desyncs* (and thus kinda breaks) these songs. So for most of the time, those songs were broken by *not* fixing them. With the introduction of this feature, those were fixed automatically. However, now I have found two counterexamples where the video actually is not in sync with its own video, requiring a #VIDEOGAP for only a single a/v source. Therefore, the automatic fix is not possible anymore, even though these cases are rare. That's why I kept the warning - telling the user that the #VIDEOGAP is *most likely* wrong (and he/she should check manually and ideally fix the song on USDB, either by finding the suitable second resource or by removing the #VIDEOGAP).

randompersona1:
I forgot the fix was only recently implemented, so it doesn't really matter. I was referring to the songs that will _currently_ be automatically fixed, but wouldn't after this change.

___


> P.S. We _could_ work around this by explicitly stating the _same_ resource for audio _and_ video for these rare cases and then _not_ delete the #VIDEOGAP. Maybe this is even the better, more elegant solution.
> 
>   * `a=` only + `#VIDEOGAP`: doesn't hurt, but makes no sense either --> delete #VIDEOGAP
> 
>   * `v=` only + `#VIDEOGAP`: **desyncs** audio and video from same source --> most likely wrong --> delete #VIDEOGAP
> 
>   * `a=` and `v=` with different source + `#VIDEOGAP`: syncs audio and video from different source --> keep #VIDEOGAP
> 
>   * `a=` and `v=` with **same** source + `#VIDEOGAP`: deliberate to sync video to its own audio track --> keep #VIDEOGAP (workaround to keep #VIDEOGAP even though `a=` and `v=` point to the same source by explicitly stating both)

I like the explicitness, but "hidden" features like these are hard to document. I would wager that few people have read the relevant wiki page, and almost nobody would check back regularly for changes. 

In the sake of pushing this truly tiny and likely inconsequential change through though, I would say we do it as you described in the comment, update the songs it applies to, update the wiki, and if somebody asks, tell them how it works. For anything larger than this, we should find a better solution.

bohning:
Sounds good. So far I only know of two videos that have async a/v, so it seems a very rare case.

RumovZ (commented):
That's a lot of text for every song with VIDEOGAP. I would decrease each message's log level.

PR #426, opened 2025-06-07
Trash songs that were deleted in USDB.

bohning:


bohning:
Ah right, didn't think of that. Important catch indeed. What would you suggest?

RumovZ:
Not sure. Maybe you can check whether fetching the songlist succeded.

RumovZ (approved):

PR #440, opened 2025-07-13
This enables reordering of columns.

bohning:
This seems to work as expected. However, we can no longer use ResizeMode.Fixed, because that also implies that reordering is not possible. This also fixes the issue that I had with not being able to resize the last column.

### DIFF
--- a/src/usdb_syncer/gui/song_table/song_table.py
+++ b/src/usdb_syncer/gui/song_table/song_table.py
@@ -149,20 +149,23 @@ class SongTable:
                 existing_state = True
                 self._search.order = Column(header.sortIndicatorSection()).song_order()
                 self._search.descending = bool(header.sortIndicatorOrder().value)
+
+        header.setSectionsMovable(True)
+        header.setSectionsClickable(True)
+        header.setStretchLastSection(False)
+        header.setMinimumSectionSize(0)
+        header.setHighlightSections(True)
+
         for column in Column:
+            header.setSectionResizeMode(
+                column, QtWidgets.QHeaderView.ResizeMode.Interactive
+            )
+
             if size := column.fixed_size():
-                header.setSectionResizeMode(
-                    column, QtWidgets.QHeaderView.ResizeMode.Fixed
-                )
                 header.resizeSection(column, size)
-            # setting a (default) width on the last, stretching column seems to cause
-            # issues, so we set it manually on the other columns
             elif column is not max(Column):
                 if not existing_state:
                     header.resizeSection(column, DEFAULT_COLUMN_WIDTH)
-                header.setSectionResizeMode(
-                    column, QtWidgets.QHeaderView.ResizeMode.Interactive
-                )
 
     def build_custom_data_menu(self) -> None:
         if not (song := self.current_song()) or not song.sync_meta:


src/usdb_syncer/gui/song_table/song_table.py
```diff
@@ -149,20 +149,23 @@ def _setup_view(self) -> None:
                 existing_state = True
                 self._search.order = Column(header.sortIndicatorSection()).song_order()
                 self._search.descending = bool(header.sortIndicatorOrder().value)
+
+        header.setSectionsMovable(True)
+        header.setSectionsClickable(True)
+        header.setStretchLastSection(False)
+        header.setMinimumSectionSize(0)
```
RumovZ, line 156: Maybe set the minimum to what was formerly the fixed size of the narrowest columns?

src/usdb_syncer/gui/song_table/song_table.py
```diff
@@ -149,20 +149,23 @@ def _setup_view(self) -> None:
...
+        header.setHighlightSections(True)
+
         for column in Column:
+            header.setSectionResizeMode(
+                column, QtWidgets.QHeaderView.ResizeMode.Interactive
```
RumovZ, line 161: Does this still have to be called in a loop if it's the same for all columns?

src/usdb_syncer/gui/song_table/song_table.py
```diff
@@ -149,20 +149,23 @@ def _setup_view(self) -> None:
...
+            )
+
             if size := column.fixed_size():
-                header.setSectionResizeMode(
-                    column, QtWidgets.QHeaderView.ResizeMode.Fixed
-                )
                 header.resizeSection(column, size)
-            # setting a (default) width on the last, stretching column seems to cause
-            # issues, so we set it manually on the other columns
             elif column is not max(Column):
```
RumovZ, line 166: Probably `max(Column)` doesn't need to be handled specially anymore?

src/usdb_syncer/gui/song_table/song_table.py
```diff
@@ -149,20 +149,23 @@ def _setup_view(self) -> None:
                 existing_state = True
                 self._search.order = Column(header.sortIndicatorSection()).song_order()
                 self._search.descending = bool(header.sortIndicatorOrder().value)
+
+        header.setSectionsMovable(True)
+        header.setSectionsClickable(True)
+        header.setStretchLastSection(False)
+        header.setMinimumSectionSize(0)
+        header.setHighlightSections(True)
```
RumovZ, line 157: Seems useless, as either all or no columns are highlighted.

src/usdb_syncer/gui/song_table/song_table.py
```diff
@@ -149,20 +149,23 @@ def _setup_view(self) -> None:
                 existing_state = True
                 self._search.order = Column(header.sortIndicatorSection()).song_order()
                 self._search.descending = bool(header.sortIndicatorOrder().value)
+
+        header.setSectionsMovable(True)
+        header.setSectionsClickable(True)
```
RumovZ, line 154: What does this do?

RumovZ (changes requested):

src/usdb_syncer/gui/song_table/song_table.py
bohning: I thought it was required in addition to `setSectionsMovable()`, but it isn't.

src/usdb_syncer/gui/song_table/song_table.py
RumovZ: I think this has to be disabled explicitly.

RumovZ (approved):

PR #457, opened 2025-08-20
Replace appdirs with platformdirs

AlexanderS:
[platformdirs](https://github.com/tox-dev/platformdirs) is an active fork of the deprecated appdirs. See the notice on top of the README: https://github.com/ActiveState/appdirs It has essentially the same API.

Note: platformdirs was already in the `poetry.lock` file as dependency of `tox`, so only the group changed.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -13,18 +13,6 @@ files = [
 ]
 
 [[package]]
-name = "appdirs"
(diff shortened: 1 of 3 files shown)


bohning (approved):
Looks good to me. Thanks a lot.

PR #479, opened 2025-10-18
Add feature to submit local changes to USDB.

bohning:


src/usdb_syncer/usdb_scraper.py
```diff
@@ -571,3 +613,50 @@ def post_song_rating(song_id: SongId, stars: int) -> None:
...
+    if not txt_path:
+        return None
+    if not (txt := SongTxt.try_from_file(txt_path, logger)):
+        logger.error("Failed to parse local song txt, skipping upload.")
+        return None
+    if txt.notes.track_2 is not None:
+        txt.headers.title += " [DUET]"
+    txt.headers.video = str(sync_meta.meta_tags)  # reinsert meta tags
+
+    return str(txt).replace("\n", "\r\n")  # USDB requires \r\n (CRLF) line endings
```
RumovZ: This logic should be in SongTxt.

src/usdb_syncer/gui/mw.py
```diff
@@ -409,6 +415,60 @@ def _rate_in_usdb(self, stars: int) -> None:
...
+            for song in local_songs:
+                submit_local_changes(song)
+
+        def on_done(result: progress.Result) -> None:
+            pass
+
+        run_with_progress(
+            f"Submitting {num_songs} song{'s' if num_songs != 1 else ''}…",
+            task,
+            on_done,
```
RumovZ: This silently discards any exception.

src/usdb_syncer/usdb_scraper.py
```diff
@@ -102,13 +123,34 @@ def reset_session(cls) -> None:
...
+    username = match.group(2)
+
+    if rank_match := ROLE_REGEX.search(html):
+        rank = int(rank_match.group(1))
+        if rank == 4:
+            role = UserRole.ADMIN
+        elif rank == 3:
+            role = UserRole.MODERATOR
+        else:
+            role = UserRole.USER
```
RumovZ: Would move this into UserRole.

src/usdb_syncer/usdb_scraper.py
```diff
@@ -571,3 +613,50 @@ def post_song_rating(song_id: SongId, stars: int) -> None:
...
+
+    txt_path = sync_meta.txt_path()
+    if not txt_path:
+        return None
+    if not (txt := SongTxt.try_from_file(txt_path, logger)):
+        logger.error("Failed to parse local song txt, skipping upload.")
+        return None
+    if txt.notes.track_2 is not None:
+        txt.headers.title += " [DUET]"
+    txt.headers.video = str(sync_meta.meta_tags)  # reinsert meta tags
```
RumovZ: Would be nice if users could improve meta tags, too.

src/usdb_syncer/gui/resources/styles/diff.css
```diff
@@ -0,0 +1,100 @@
+<style>
```
RumovZ: Not CSS syntax.

src/usdb_syncer/meta_tags.py
```diff
@@ -113,8 +113,9 @@ def to_str(self, prefix: str) -> str:
         return _join_tags(
             _key_value_str(prefix, self.source),
             _key_value_str(f"{prefix}-rotate", self.rotate),
-            self.crop.to_str(prefix) if self.crop else None,
+            self.crop.to_str(prefix) if self.crop and prefix == "co" else None,
```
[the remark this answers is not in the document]
bohning: The canonical order for cover is (crop, resize), while for backgrounds, it's (resize, crop). While meta tag order is irrelevant for processing, I'd prefer to have it in canonical order instead of having (crop, resize) for both covers and backgrounds. Maybe there is an easier way?

src/usdb_syncer/usdb_scraper.py
bohning: True, didn't think of that. However, I guess the main use case is to download a song, then correct/improve the local file (in USDX, YASS, or even a text editor), then submit the changed file (that then does not contain the meta tags any longer). Do you have an idea how this could be implemented?

src/usdb_syncer/gui/mw.py
bohning: So what can be done about it? Run it without progress? For the normal user, it is limited to a single song anyways, so I guess this is only relevant for batch uploads by mods/admins.

src/usdb_syncer/meta_tags.py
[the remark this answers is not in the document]
bohning: Yes. Most covers are already square (Apple, fanart) and do not need any postprocessing. Some may require cropping to square dimensions. Rarely, it is impossible to crop non-square covers to square dimensions, then they should be cropped "as square as possible without cropping important parts", then resize to square dimensions to keep distortions as small as possible (hence: first crop, then resize to square dimensions).

For backgrounds, if they are not 16:9, it's easier to resize first to have either height or width fit the target dimension, then crop in the other dimension to get a 16:9 image (hence: first resize, then crop).

This is already done like when postprocessing and the order of the meta tags is irrelevant for the postprocessing by the Syncer, but I'd prefer to have them in canonical order co-crop,co-resize,bg-resize,bg-crop.

src/usdb_syncer/gui/mw.py
```diff
@@ -398,6 +406,54 @@ def _rate_in_usdb(self, stars: int) -> None:
...
+            if current_song := self.table.current_song():
+                selected_songs.append(current_song)
+        else:
+            selected_songs = list(self.table.selected_songs())
+
+        if not selected_songs:
+            QMessageBox.information(
+                self, "No songs selected", "No songs selected for submission."
+            )
+            return
```
RumovZ: Can be removed.

src/usdb_syncer/usdb_scraper.py
```diff
@@ -26,6 +27,35 @@
...
+        return self.value
+
+
+@dataclass(frozen=True)
+class UsdbUser:
+    name: str
+    role: UserRole
+
+    @classmethod
+    def from_rank(cls, name: str, rank: int | None) -> "UsdbUser":
```
RumovZ: `from __futures__ import annotations` so you don't have to use quotes.

src/usdb_syncer/gui/usdb_upload_dialog.py
```diff
@@ -0,0 +1,108 @@
...
+from usdb_syncer.gui import progress, theme
+from usdb_syncer.gui.forms.UsdbUploadDialog import Ui_Dialog
+from usdb_syncer.gui.resources import styles
+from usdb_syncer.gui.theme import DiffPalette, _rgb_str
+from usdb_syncer.logger import song_logger
+from usdb_syncer.usdb_scraper import submit_local_changes
+from usdb_syncer.usdb_song import SongChanges, UsdbSong
+
+
+def _html_prefix(action: str = "", symbol: str = " ") -> str:
```
RumovZ: unused (should also be catched by CI)

src/usdb_syncer/gui/usdb_upload_dialog.py
```diff
@@ -0,0 +1,108 @@
+"""Dialog to manage USDB uploads."""
+
+from __future__ import annotations
+
+from PySide6.QtWidgets import QDialog, QWidget
+
+from usdb_syncer.gui import progress, theme
+from usdb_syncer.gui.forms.UsdbUploadDialog import Ui_Dialog
+from usdb_syncer.gui.resources import styles
+from usdb_syncer.gui.theme import DiffPalette, _rgb_str
```
RumovZ: Private import (wonder why ruff allows that). Maybe move generate_diff_css into theme.py.

src/usdb_syncer/usdb_song.py
```diff
@@ -227,6 +232,156 @@ def set_playing(self, is_playing: bool) -> None:
...
+    remote_lines: list[str] = attrs.field(factory=list)
+    local_lines: list[str] = attrs.field(factory=list)
+    line_num_remote: int = 1
+    line_num_local: int = 1
+    left_line: str = ""
+    right_line: str = ""
+    left_has_change: bool = False
+    right_has_change: bool = False
+
+    def add_content(self, part: str, op: int, renderer: Callable) -> None:
```
RumovZ: Type annotation should contain parameters and return type of callable.

RumovZ:
> Yes, the code differentiates between insertions and deletions, but in the end I thought it would be cleaner not differentiate between them in the diff. I left it this way so this could easily be changed in future.

Yeah, I guess different colours are only really required if you have a unified instead of a split diff view.

Another question: Should line 11 in the local view not be highlighted here? Also, line wrapping causes misalignment, so you probably would want to change the overflow behaviour.

image

PR #489, opened 2025-11-04
Fallbacks

bohning:
image

### DIFF
--- a/src/usdb_syncer/song_loader.py
+++ b/src/usdb_syncer/song_loader.py
@@ -743,34 +743,23 @@ def _write_headers(ctx: _Context) -> None:
     if version >= FormatVersion.V1_1_0:
         ctx.txt.headers.providedby = constants.Usdb.BASE_URL
 
-    if path := ctx.out.audio.path(ctx.locations, temp=True):
-        _set_audio_headers(ctx, version, path)
-    else:
+    _set_audio_headers(ctx, version)
+    _set_video_headers(ctx, version)
+    _set_cover_headers(ctx, version)
+    _set_background_headers(ctx, version)
+
+
+def _set_audio_headers(ctx: _Context, version: FormatVersion) -> None:
+    path = ctx.out.audio.path(ctx.locations, temp=True)
+
+    if not path:
         ctx.txt.headers.mp3 = None
         ctx.txt.headers.audio = None
         ctx.txt.headers.audiourl = None
+        return
 
-    if path := ctx.out.video.path(ctx.locations, temp=True):
-        _set_video_headers(ctx, version, path)
-    else:
-        ctx.txt.headers.video = None
-        ctx.txt.headers.videourl = None
-
-    if path := ctx.out.cover.path(ctx.locations, temp=True):
-        _set_cover_headers(ctx, version, path)
-    else:
-        ctx.txt.headers.cover = None
-        ctx.txt.headers.coverurl = None
-
-    if path := ctx.out.background.path(ctx.locations, temp=True):
-        _set_background_headers(ctx, version, path)
-    else:
-        ctx.txt.headers.background = None
-        ctx.txt.headers.backgroundurl = None
-
-
-def _set_audio_headers(ctx: _Context, version: FormatVersion, path: Path) -> None:
     fname = ctx.locations.filename(ext=utils.resource_file_ending(path.name))
+
     match version:
         case FormatVersion.V1_0_0:
             ctx.txt.headers.mp3 = fname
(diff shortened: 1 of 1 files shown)


src/usdb_syncer/utils.py
```diff
@@ -462,3 +463,15 @@ def __exit__(
 
             os.environ.clear()
             os.environ.update(self.modified_env)
+
+
+class JobResult(StrEnum):
```
RumovZ: Not really a utility. Would put it in the sql module.

src/usdb_syncer/sync_meta.py
```diff
@@ -33,40 +33,56 @@ def __str__(self) -> str:
 class ResourceFile:
     """Meta data about a local file."""
 
-    fname: str
-    mtime: int
-    resource: str
+    fname: str | None
+    mtime: int | None
+    resource: str | None
```
RumovZ: This is unfortunately allows many invalid configurations. Better to have a wrapper witht two fields, JobResult and the old ResourceFile.

src/usdb_syncer/db/__init__.py
```diff
@@ -270,33 +271,65 @@ def sql(self) -> str | None:  # noqa: C901
...
+                return (
+                    "CASE "
+                    "WHEN txt.sync_meta_id IS NULL THEN 0 "
+                    "WHEN txt.status = 'skipped_disabled' THEN 1 "
+                    "WHEN txt.status = 'skipped_unavailable' THEN 2 "
+                    "WHEN txt.status = 'failed' THEN 3 "
+                    "WHEN txt.status = 'failed_existing' THEN 4 "
+                    "WHEN txt.status = 'fallback' THEN 5 "
+                    "WHEN txt.status = 'success' THEN 6 "
+                    "ELSE 7 END"
```
RumovZ: Is repeated 4 times instead of using a single f string.
Alternatively, the status could also be stored as a number.

src/usdb_syncer/utils.py
bohning: In `db.sql.__init__.py`? Or rather `sql.__init__.py`?

src/usdb_syncer/gui/resources/qt/success-white.svg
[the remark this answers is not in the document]
bohning: Yeah, thought about that but then thought to harmonize with the newly added icons. I can go back to descriptive names and map them to semantic names in icons.py.

src/usdb_syncer/song_loader.py
```diff
@@ -421,8 +422,26 @@ def _maybe_download_audio(ctx: _Context) -> JobResult:
...
+            if (
+                ctx.txt.meta_tags.is_audio_only()
+                and (resource == ctx.txt.meta_tags.audio)
+            ) or (
+                not (ctx.txt.meta_tags.is_audio_only())
+                and (
+                    resource == ctx.txt.meta_tags.audio
+                    or resource == ctx.txt.meta_tags.video
+                )
+            ):
```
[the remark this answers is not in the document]
bohning: Do you mean `def all_audio_resources(self) -> Iterator[AudioResource]:` with an `AudioResource` dataclass that contains the URL and a bool to indicate if it's a fallback?

src/usdb_syncer/db/__init__.py
```diff
@@ -271,69 +271,44 @@ def sql(self) -> str | None:  # noqa: C901
...
-                    "ELSE 7 END"
-                )
+                return _status_case_sql("background")
             case SongOrder.STATUS:
                 return _STATUS_COLUMN
 
 
+def _status_case_sql(alias: str) -> str:
+    """Return CASE SQL expression for resource_file status sorting."""
+    statuses = [
```
RumovZ: I prefer to define this on the class itself, like a method `order()`.

src/usdb_syncer/gui/song_table/song_table.py
```diff
@@ -218,8 +218,13 @@ def _play_or_stop_sample(self, song: UsdbSong) -> None:
             self._media_player.stop()
             return
         position = 0
-        if song.sync_meta and song.sync_meta.audio and song.sync_meta.audio.fname:
-            path = song.sync_meta.path.parent / song.sync_meta.audio.fname
+        if (
+            song.sync_meta
+            and song.sync_meta.audio
+            and song.sync_meta.audio.file
+            and song.sync_meta.audio.file.fname
```
RumovZ: There's already the `audio_path()` helper for this.

src/usdb_syncer/sync_meta.py
```diff
@@ -33,72 +33,107 @@ def __str__(self) -> str:
...
+
+    @classmethod
+    def new(cls, status: JobStatus, file: ResourceFile) -> Resource:
+        """Create a successful status with a file."""
+        return cls(status=status, file=file)
+
+    @classmethod
+    def status_only(cls, status: JobStatus) -> Resource:
+        """Create a failed/skipped status without a file."""
+        return cls(status=status, file=None)
```
RumovZ: These are now redundant as you can just call `Resource(status, file)`.

CHANGELOG.md
RumovZ: The yt-dlp fix and deno requirement should be mentioned conspicuously.

src/usdb_syncer/song_loader.py
```diff
@@ -745,15 +745,28 @@ def _write_headers(ctx: _Context) -> None:
 
     if path := ctx.out.audio.path(ctx.locations, temp=True):
         _set_audio_headers(ctx, version, path)
+    else:
+        ctx.txt.headers.mp3 = None
```
RumovZ: I'd find it more comprehensible if audio headers were only set in a single place. So e.g. `_set_audio_headers` should also handle the `None` case. It also doesn't need to be passed `path`, which is derived from `ctx`.

PR #510, opened 2026-01-18
Fix imports creating a QPixmap, leading to errors

randompersona1:
Importing mw.py no longer leads to an error if a QtGuiApplication doesn't exist yet.

This allows addons to continue importing mw.py normally. After #504, they would raise an error: "QPixmap: Must construct a QGuiApplication before a QPixmap".


randompersona1:
Would this achieve what you're imagining? It's a little more boilerplate, but I think it is otherwise equal to the previous constant (except it isn't initialized on import).

PR #525, opened 2026-01-29
Testing the GUI on linux distros

randompersona1:
The new `linux_distr_testing` workflow builds a binary and a wheel, then runs a healthcheck on various linux distros. I chose to install the display dependencies for X11 instead of wayland, because I thought it would be more compatible. It should be possible to use `wlheadless-run` to do the same thing for wayland, but doing both seemed overkill to me, since it's very unlikely we'd find any errors.

Since I went to the trouble of figuring out which packages we need for various distros anyways, I updated the readme for more clarity and better instructions regarding linux.

The PR also includes a properly working testpypi upload in case we want to verify the wheels manually.


src/usdb_syncer/gui/__init__.py
```diff
@@ -89,6 +90,9 @@ def parse(cls) -> CliArgs:
         dev_options.add_argument(
             "--profile", action="store_true", help="Run with profiling."
         )
+        dev_options.add_argument(
+            "--healthcheck", action="store_true", help="Run healthcheck and exit."
+        )
```
RumovZ: This actually seems more like a subcommand than a flag.

pyproject.toml
```diff
@@ -69,24 +69,15 @@ default-groups = [
 source = "vcs"
 
 [tool.hatch.build.targets.sdist]
-include = [
+artifacts = [
```
RumovZ: I think this will bundle some redundant stuff like UI files and media already bundled by Qt.

README.md
```diff
@@ -114,36 +120,58 @@ startup phase, they cannot be completely ruled out.
...
 
-Linux bundles are generated on AlmaLinux 9. They should be compatible with any modern distribution. If not, please open an issue.
+</details>
 
-The only known requirement for the binary is `glibc >= 2.34`. The current `glibc` version can be checked with:
+You can also use the corresponding wayland libraries. We will fix issues related to wayland, but we are unable to efficiently test on wayland properly, which is why problems may be more frequent.
+
+### Compatibility
+
+The wheels are pure python with no extension modules, and are thus only restricted by the python version.
```
RumovZ:
```suggestion
The wheels are pure Python with no extension modules, and are thus only restricted by the Python version.
```

pyproject.toml
RumovZ: True. This was probably written under the assumption that like with PyInstaller, non-Python files were excluded by default. Can we then exclude the files currently bundled twice explicitly?

RumovZ:
Is it very cumbersome keeping the full release as it is? If yes, I'm fine with changing it. Otherwise, I would like to keep it last, because if any other step does fail and the PyPI upload completed, it could become rather difficult to fix up the release.

randompersona1:
Okay, now we have:

- `dev_pypi.yaml`, which runs when a commit without a tag is pushed to main and builds to pypi, stripping the local part first
- `testpypi.yaml`, which can be run manually and pushes to testpypi, stripping the local part first
- `release.yaml`, which still uploads to pypi, but only runs in the first place if a tag is pushed, meaning it will always have a full pep440 version
- `linux_distro_testing.yaml`, which doesn't upload anything, but verifies that both wheels and bundles are built correctly and work on a wide range of linux distros with minimal dependencies

PR #531, opened 2026-01-31
Fix merged Noto font.

bohning:
This one was created like the previous one, but in reverse order (so merging the regular font into the symbol font).


randompersona1:
I can confirm this file works where the previous font file failed. I haven't done super broad testing, but I'm fine with merging regardless.

You may want to include a changelog entry.

randompersona1 (approved):

Issue #517, opened 2026-01-26
Status column gone after update


Nuhser:
The status column which I had placed on the left side of the table is gone entirely after the update from release 0.18.0 to 0.19.0. Instead, the column for the downloaded background image moved from the far right side of my table to the left side where the status column used to be.

I can't seem to find either the status column itself somewhere in the table or a way to enable it again.

**OS:** Fedora Linux 43
**Desktop Environment:** KDE Plasma 6.5.4
**USDB Syncer Version:** 0.19.0

PR #537, opened 2026-02-02
Add column for custom data.

bohning:
Icon indicates presence of custom data, tooltip lists key-value pairs.
image

Not sure how the sorting could be improved.


RumovZ:
I'll look into configurable custom columns. Let's come back to this PR afterwards to see if still makes sense.

> I agree it's unfortunate. If there's a remedy, I am all for it. Otherwise I find it somewhat acceptable.

I find it terrible, but I realised it might actually be a bug. For me, all columns get reduced to the minimal size (`MIN_COLUMN_WIDTH = 24`), even though we have code to set it to a more reasonable default (`DEFAULT_COLUMN_WIDTH = 300`). See the screenshot in #517.
This makes the table unusable until I have manually expanded every single column (apart from the icon ones).
Does this work for you? Anyway, I'll look into that, too.

PR #541, opened 2026-02-03
Fix sample

randompersona1:
Sample was broken by #524. I searched a little, and I think this is the only place with that bug.

Bit of a spotlight on how poorly our GUI modules are tested. Very difficult to spot in a code review, and impossible to check for analyzer tools or even the healthcheck.

Edit: tbf, I've thought some more, and I don't see how any manual test case would have caught this either. Which raises the question on how to prevent similar bugs in the future?


bohning:
Good catch! But only clicking was broken, hitting space to play/pause still works (and as a Mac user, that's the most natural thing to do for a preview).

PR #543, opened 2026-02-05
Fix memory leaks in Previewer and dialogs

RumovZ:


randompersona1:
Why exactly is this needed? When the python object goes out of scope (or maybe only when it's garbage collected, not sure), shiboken also removes the corresponding C++ object. What stays alive? 

More importantly, what happens if the window is closed *before* the python object is destroyed? Then we'll essentially be holding a nullptr, no? I'm unsure how shiboken will handle that.

Can we verify this behaviour works somehow? Have you observed memory leaks currently?

bohning (approved):

randompersona1 (approved):

PR #546, opened 2026-02-06
Proper env cleaning

### LINKED ISSUE #545: Implement LinuxEnvCleaner properly
Currently, `utils.LinuxEnvCleaner` is fragile. It's actually surprising I haven't seen it cause problems yet. Modifying `os.environ` across multiple different threads at the same time is very bad practice.

Ideally, we wouldn't modify `os.environ` at all, and pass the `env` argument to subprocess.run() and subprocess.Popen(). `ffmpeg-normalize` supports this from 1.37 onwards.

I think the main reason it works currently is that all of our modules load dynamic libraries at import time, when the pyinstaller environment is untouched. But that is obviously implementation dependant, and I can easily see how a dependency could introduce lazy loading at some point.

Looking at sounddevice and webbrowser, which are the two problematic ones, it may be possible to just monkeypatch in the correct binary/environment.

---------------------

For clarity, the problem the class solves is that in a pyinstaller bundle, pyinstaller modifies `LD_LIBRARY_PATH` so that the bundled libraries are used instead of the system ones. So for example, our bundles use the bundled sqlite regardless of whether it's installed on the user system or not.

But when we call a process like ffmpeg, it inherits the environment we set. So when ffmpeg tries to load e.g a video library that Qt *also* uses and is therefore bundled, it sees the bundled library first, and there's a version mismatch.

randompersona1:
Resolves #545

Changes to env cleaning:
- monkey patch webbrowser to use system environment. webbrowser is stable and I don't anticipate changes, so this is okay.
- load sounddevice modifying the global environment. I tried similar approaches as for webbrowser, but it's just hell. First import of sounddevice needs to be controlled, this currently happens in the previewer. Once portaudio is loaded into memory, we don't need to modify the environment for further calls.
- for ffmpeg-normalize, use native method introduced in 1.37.0
- for our own subprocess.run/Popen, use env=subprocessing.get_env_clean() when accessing system cli tools

These changes pass the automated tests, and I've verified it works on a fedora system.

Heathcheck now also errors when logger.exception was called.

Changes should be stable in a nogil environment as well, whereas the current implementation is likely highly unsafe.

### DIFF
--- a/src/usdb_syncer/gui/__init__.py
+++ b/src/usdb_syncer/gui/__init__.py
@@ -2,7 +2,9 @@
 
 from __future__ import annotations
 
+import contextlib
 import cProfile
+import io
 import logging
 import shutil
 import subprocess
@@ -141,7 +143,7 @@ class CliArgs:
 
     def apply(self) -> None:
         if self.healthcheck:
-            sys.exit(_run_heathcheck())
+            sys.exit(_run_healthcheck())
         if self.reset_settings:
             settings.reset()
             print("Settings reset to default.")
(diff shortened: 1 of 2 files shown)


src/usdb_syncer/logger.py
```diff
@@ -36,6 +39,8 @@ def exception(
             self.debug(None, exc_info=exc_info, **kwargs)
         if msg:
             self.error(msg, *args, exc_info=False, **kwargs)
+        if _SHUTDOWN_ON_ERROR:
+            sys.exit(2)
```
RumovZ: If find it surprising for a logger to call sys exit. The usual approach is to check if anything was sent to stderr.

src/usdb_syncer/subprocessing.py
```diff
@@ -0,0 +1,86 @@
...
+    os.environ.update(safe_env)
+    try:
+        yield
+    finally:
+        os.environ.clear()
+        os.environ.update(original_env)
+        environ_lock.release()
+
+
+def patch_webbrowser_subprocess() -> None:
```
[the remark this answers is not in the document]
randompersona1: Not easily, since there is no bundled `python.exe`. Besides, spinning up the entire interpreter is a lot of overhead. 

I suppose we could bundle a tiny rust binary or something and call that.

src/usdb_syncer/logger.py
RumovZ: Still, the kill switch makes me a bit nervous. 😅
I'd just grep stderr for `[ERROR]`.

randompersona1: That's not so easy if you don't know bash well, and I don't. Gemini is giving me a pretty complicated command.

I'll remove the kill switch and we can come back another time. Doesn't matter for the rest of the PR anyways.

RumovZ (approved):

PR #547, opened 2026-02-07
Logging improvements

randompersona1:
Cleaned up the log and especially the terminal output. Users should use the log file if they need to share the log instead of copying from the terminal.

Qt messages are suppressed, at least those that listen. This can be re-enabled using `--show-debug`, which may be helpful for diagnosing issues. Sadly, the song sample and other multimedia stuff still writes a bunch of junk to stderr, and I'm not prepared to remove that just yet.

### DIFF
--- a/src/usdb_syncer/logger.py
+++ b/src/usdb_syncer/logger.py
@@ -79,6 +79,7 @@ GUI_FORMATTER = logging.Formatter(
 DEBUG_FORMATTER = logging.Formatter(
     style="{",
     fmt="{asctime} [{levelname}] [{filename}:{lineno}, {threadName}] {message}",
+    datefmt="%Y-%m-%d %H:%M:%S",
 )
 
 


src/usdb_syncer/usdb_scraper.py
```diff
@@ -316,10 +316,8 @@ def _get_usdb_page_inner(
     url = Usdb.BASE_URL + rel_url
     match method:
         case RequestMethod.GET:
-            logger.debug(f"Get request for {url}")
```
RumovZ: I think network traffic is one of the areas where rigorous logging is warranted. Lot's of stuff that could go wrong there.

randompersona1: Fair enough, but I don't think those two statements in particular do anything special. I don't need:

> Post request for https://usdb.animux.de/index.php

This occurs on startup, and especially when synchronizing the song list. Way too little detail to be helpful, and there's no follow-up log message that would help us diagnose an issue. If an exception does occur, which it will if the connection fails, that will show up in the log anyways.

RumovZ:
I thought there was some technical reason the PyInstaller console was enabled at the time, but maybe I misremember. From a user perspective, I think it is completely redundant and to some even confusing.

randompersona1:
Both portable and install versions work fine for me. I suspect users will find it initially more difficult to share the log, since many just copy the console output, but I can live with that.

RumovZ:
> IMO, stdout would be significantly more useful if a lot of boilerplate was removed.

IMO, a log is bad for giving user feedback in general. We're cramming three very different use cases into it: Feedback for user actions, stats for USDB maintainers (i.e. bohning), and debugging info for developers. The log currently works more or less for the last two, but to the detriment of the first one.

We have already taken some steps to reduce reliance on the log, but there remains a lot of work to be done before we can hide the log by default for regular users.

RumovZ:
> Are you talking about hiding the console with the pyinstaller cli flag, or the default option for "--log-level" being "debug", or the "--show-debug" flag?

Yeah, sorry, 'log' was rather vague. I was talking about hiding the log widget in the UI.
Regardless of the message log level, communicating anything to users via an ever growing wall of text is not ideal.

I'm aware this is not the focus of this PR. It was just to explain why I'm less concerned with the exact messages we currently log: In the long run we should get rid of any user communication therein anyway, and then we're free to treat it more like a developer tool.

randompersona1:
Absolutely. This PR is meant as an incremental improvement to the stdout log, and it will hopefully outlive the log window in the UI. The impulse to change it comes from me being annoyed at seeing the same messages every time I run the syncer from source. I'm fully aware of the need to rethink user communication entirely.

src/usdb_syncer/gui/__init__.py
```diff
@@ -157,14 +163,16 @@ def apply(self) -> None:
 
 
 def main() -> None:
-    if hasattr(sys, "_is_gil_enabled") and sys._is_gil_enabled() is False:
+    sys.excepthook = _excepthook
+    if hasattr(sys, "_is_gil_enabled") and sys._is_gil_enabled() is False:  # type: ignore[attr-defined]
```
RumovZ:
```suggestion
    if getattr(sys, "_is_gil_enabled", False):
```

src/usdb_syncer/logger.py
```diff
@@ -32,12 +51,23 @@ def exception(
         **kwargs: Any,
     ) -> None:
         """Log exception info with debug and message with error level."""
+        print(msg, args, exc_info)
```
RumovZ: Deliberate?

src/usdb_syncer/gui/__init__.py
randompersona1: `_is_gil_enabled` is a function. getattr doesn't call it. But if you have a different idea to get rid of the `type: ignore`, I'm all ears.

RumovZ: Ah, that's what I've missed. Maybe `getattr(sys, "_is_gil_enabled", lambda: False)()` works?

RumovZ (approved):
Much nicer, thanks!

PR #567, opened 2026-03-10
Reworked main entry point

randompersona1:
Separated main entrypoint from gui module.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,7 +37,7 @@ dependencies = [
 Repository = "https://github.com/bohning/usdb_syncer"
 
 [project.scripts]
-usdb_syncer = "usdb_syncer.main:main"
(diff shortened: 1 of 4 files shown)


RumovZ:
Not sure how removing PySide6 completely would work, since we're also using it for stuff like threading and events. But yeah, that would be ideal.
Of course, it doesn't have to be done in this PR. But if this is the direction we want to go in, it would make sense to move the entry point to a `gui` module.

randompersona1:
How would you have the pypi namespace work? Have a usdb_syncer_gui package? Or a usdb_syncer_cli package, with usdb_syncer including the Gui?

RumovZ:
The latter. `usdb_syncer` should remain the go-to package, with `usdb_syncer_cli` or `usdb_syncer_core` as a dependency and stand-alone package for advanced use cases.

Issue #552, opened 2026-02-17
Getting rid of log-based user communication


RumovZ:
As discussed previously, last in #547, the log widget should not and cannot be used to reliably communicate information to the user. This issue is to track instances where log messages are currently used in such a way, and find alternatives.
Feel free to edit.

<hr>

I've identified three categories of events that I would give user feedback for, where there's currently only a log message. Anything related to USDB and resource quality remaining solely in the log is fine with me.

## Failed synchronous actions

GUI alternatives: toast, modal dialog.

- [ ] Not logged in, skipping song submission.
- [ ] Song does not exist locally anymore.
- [ ] Song does not exist locally.
- [ ] No current song.
- [ ] no files selected to import USDB IDs from
- [ ] Skipping export: no songs selected.
- [ ] export aborted
- [ ] Cannot submit: song is not remote.
- [ ] Cannot submit: remote song parsing failed.
- [ ] Cannot submit: song is not local.
- [ ] Cannot submit: song is not synchronized.
- [ ] Cannot submit: song has no local changes.
- [ ] Not downloading song as it is pinned.
- [ ] Not trashing song folder as it is pinned.
- [ ] imported USDB IDs are not available:
- [ ] Failed to import file '{path}': {error!s}
- [ ] Failed to fetch new songs; check network connection.

## Successful synchronous actions

GUI alternatives: toast.

- [ ] Local changes successfully submitted to USDB.
- [ ] exported {len(selected_ids)} USDB IDs to {path}
- [ ] PDF report created at {path}.
- [ ] JSON report created at {path} ({num_of_songs} songs).
- [ ] Webserver is now running on {address()}
- [ ] Webserver was stopped.
- [ ] available \_ imported USDB IDs will be selected:
- [ ] Fetched \_ updated song(s) from USDB

## Background errors

Here, something more involved is required: Maybe some sort of indicator that something went wrong, a searchable / filterable overview of errors, and stopping further downloads for persistent errors (like missing USDB login).

- [ ] _Baiscally everything in `song_loader.py` and `usdb_scraper.py`._

RumovZ:
I've now switched to py-spy. It confirms that a lot of time is spent logging. However, contrarily to what I've written above, it doesn't seem to matter whether the main window is already shown or not. No idea why it sometimes takes 30 and sometimes 200 seconds. Very weird. 😕
[...]

PR #568, opened 2026-03-13
Toast implementation

randompersona1:
An implementation for #552 

Haven't added usage yet, so you'll have to test using the REPL (using `ToastManager.show_message()`). 

The `paintEvent` override is a GPT suggestion. I couldn't get the stylesheet to render otherwise (could be a windows thing). I'd be glad for a different solution.

Also, it would be possible to display the toast over every other window, which could be seen as intrusive, but also makes sense for a notification system.

### DIFF
--- a/src/usdb_syncer/gui/mw.py
+++ b/src/usdb_syncer/gui/mw.py
@@ -25,7 +25,6 @@ from usdb_syncer.gui.debug_console import DebugConsole
 from usdb_syncer.gui.forms.MainWindow import Ui_MainWindow
 from usdb_syncer.gui.licenses_dialog import LicensesDialog
 from usdb_syncer.gui.meta_tags_dialog import MetaTagsDialog
-from usdb_syncer.gui.notification import ToastManager
 from usdb_syncer.gui.previewer import Previewer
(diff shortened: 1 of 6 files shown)


randompersona1:
I cheated a little by saving a MainWindow instance. But it seems pretty ergonomic to me. The previewer standalone window can store itself there also.

I played around a little with the colors, and I'm reasonably happy now. I think there is some caching that can be done, but I need more testing.

My main concern is if we want some kind of record for toasts in case the user misses one.

RumovZ:
I just noticed, "Fetched _ updated songs from USDB" is emitted before the main window is shown, so this will be tricky either way. I assume this is an exception and not the rule, and would be fine with skipping the toast here.
The idea was mostly to add the toast for synchronous user actions. I.e. the user clicks a button and a message appears. In these cases, calling the toast manager should be perfectly straightforward.
I wouldn't add a new handler system unless we really need it. For edge cases, events could be used. Performance is not an issue unless you have thousands of events per second, but such a number of toasts would break the GUI anyway.

RumovZ:
From my side, this is good to go. I think it already covers the most important use cases.

What do you think?

PR #572, opened 2026-03-31
Stem Separation

### LINKED ISSUE #168: Option to automatically create a karaoke/instrumental version by stripping vocals from audio
Spleeter or demucsV4 do a pretty good job in separating vocals and instrumental part and could be used for that.

Pro:
- real karaoke without the original singer’s voice

Con:
- will take long without GPU/CUDA
- will bloat the executable

randompersona1:
Closes #168 

Much of the PR is based on or directly copied from #523 (thanks again, @bohning )

Also relevant is https://github.com/bohning/usdb_syncer_separation/blob/main/SPEC.md and https://github.com/bohning/usdb_syncer/wiki/Separation-Providers (both not finished entirely)

-------------------------------------------

The primary architectural question was how to handle the separation binary: keep it open centrally in a global, or open it multiple times. I chose for every thread to have to open it separately to allow changing settings mid-download and having it handled like every other setting.

The current implementation deliberately trusts the separation binary to work. For example, there are no timeouts, little process killing, little error handling. I experimented a little with those things and plan to improve this in the future, but it would have caused bloat here and delayed the feature.

Another idea I had was a window to show stderr of the separation binary (difficult for downloads anyways, requires #552 ).

I didn't write a test for the JSON-RPC client because it is so simple, but I could if you want.


src/usdb_syncer/gui/resources/qt/tick-parentheses-white.svg
RumovZ, line 1: Seems this branch contains some outdated files.

src/usdb_syncer/gui/settings_dialog.py
```diff
@@ -146,6 +214,50 @@ def _get_executable(self, app: settings.SupportedApps) -> Path | None:
...
 
+    def _select_separation_executable(self) -> Path | None:
+        filt = "*"
+        directory = platformdirs.user_desktop_dir()
+        filename = QFileDialog.getOpenFileName(
+            self, "Select separation provider", directory, filt
+        )[0]
+        self.raise_()
+        if not filename:
+            notification.error("No file selected.")
```
RumovZ, line 225: I think in cases like these where a user actively cancels a dialog, no notification is needed.

src/usdb_syncer/gui/settings_dialog.py
```diff
@@ -146,6 +214,50 @@ def _get_executable(self, app: settings.SupportedApps) -> Path | None:
...
+            self,
+            f"{name}",
+            f"{name} v{self._separation_manager.get_version()}\n"
+            f"Location: {self._separation_manager.client.command[0]}\n"
+            f"Available Models: {', '.join(self._separation_manager.get_available_models())}",
+        )
+
+    def _update_provider_label(self, state: _ProviderState) -> None:
+        self._provider_state = state
+        match state:
```
RumovZ, line 243: Should be marked as exhaustive.

src/usdb_syncer/separation/client.py
```diff
@@ -0,0 +1,112 @@
...
+                future.set_exception(CommunicationError(msg))
+
+    def request(self, method: str, params: dict | list | None = None) -> Any:
+        if not self._process or not self._process.stdin:
+            msg = "Client not started"
+            raise CommunicationError(msg)
+
+        with self._pending_lock:
+            request_id = next(self._id_generator)
+            future: concurrent.futures.Future = concurrent.futures.Future()
```
RumovZ, line 91: Redundant annotation?

src/usdb_syncer/separation/separation.py
```diff
@@ -0,0 +1,126 @@
...
+def set_max_concurrent(n: int) -> None:
+    """Update the global concurrency limit for split operations."""
+    _semaphore.resize(n)
+
+
+class SeparationManager:
+    """Manages stem separation."""
+
+    def __init__(
+        self, command: list[str], logger: logger.Logger = logger.logger.logger
```
RumovZ, line 68: Why `logger.logger.logger`?

src/usdb_syncer/song_loader.py
```diff
@@ -635,6 +652,105 @@ def _try_download_audio_or_video(
     return JobStatus.FAILURE
 
 
+def _maybe_separate_stems(ctx: _Context) -> JobStatus:
+    if not ctx.options.audio_options or not ctx.options.audio_options.separation:
+        ctx.logger.info("Stem separation is disabled, skipping.")
```
RumovZ, line 657: A bit noisy, given that this will remain the overwhelming default.

RumovZ (commented):
- The table header is broken. (Could be an existing bug that you've triggered by adding new columns.)
- The buttons in the settings dialog are a bit confusing. No feedback is provided after selecting a (bad) provider, and it's not clear what `Info` does.
- Then there's another info button with an icon. This should have a dark variant.
- I think there should be an icon indicating the state of the selected separator? Seems to only partially work.
- The outer padding of the settings dialog has disappeared.

randompersona1:
> The table header is broken. (Could be an existing bug that you've triggered by adding new columns.)

I can't reproduce this. The header is fine for me, and @bohning didn't mention any issues either. Could this be something stateful, e.g depending on the saved geometry in the settings?

> I think there should be an icon indicating the state of the selected separator? Seems to only partially work.

Could you elaborate? The icon and its tooltip should change based on separation provider state.

> The buttons in the settings dialog are a bit confusing. No feedback is provided after selecting a (bad) provider, and it's not clear what Info does.

I don't disagree. I've spent quite a lot of time thinking about this, but can't come up with something I truly like. Do you have some general suggestions?

The settings we need are:
- A file selector button to select a provider
- Some sort of state communication to see that a provider is selected
- A combobox to select a model from the provider
- A spinbox to choose the number of simultanious separations
- An Info button that links to the wiki page (this would be really nice in the groupbox header, but I can't get it there)
- (kind of optionally) a different Info button that displays more detailed information about the separation provider

RumovZ:
> Could this be something stateful, e.g depending on the saved geometry in the settings?

Maybe because I have a custom column? This is how it looks:

image

> The icon and its tooltip should change based on separation provider state.

When I choose an invalid provider (don't have a valid one to test with), the icon disappears:

image

> Do you have some general suggestions?

I think it's a good idea to display the path to the selected provided, like in App Paths. And when the selected provider is invalid, I would show an error and reset it immediately. This makes the warning icon redundant.
The other elements should be disabled when no provider is selected.

If it's not possible to have the help button in the group box title, I would at least put it somewhere at the top, above the selector. And it could be a bit more expressive, maybe a label "Learn more ..." or something like that.

src/usdb_syncer/separation/client.py
randompersona1: No, mypy requires it for some reason.

randompersona1:
> Maybe because I have a custom column? This is how it looks:

I see. I can reproduce, but I don't really understand the issue (The header should be resetting, but it doesn't and tries to apply the old geometry). It fixes itself for me by just closing and reopening the syncer though.

> When I choose an invalid provider (don't have a valid one to test with), the icon disappears:

In case you want to, check out https://github.com/bohning/usdb_syncer_separation , it's pretty simple. I fixed the error though.

I also managed to get the icon to the top of the groupbox. I can't put any text in that label, the paddings would be way too big and interfere with the content above, but IMO the icon is okay.

> I think it's a good idea to display the path to the selected provided, like in App Paths

It's shown in the "Provider Info" dialog. I put it there because we don't really have enough space to show a full-length path.

I somehow broke the width of the settings dialog, so I'll have to see how that goes back, but otherwise, I'm quite happy now.

src/usdb_syncer/song_loader.py
randompersona1: Made it a debug message

src/usdb_syncer/gui/resources/qt/tick-parentheses-white.svg
randompersona1: Removed, and included a test so it doesn't happen again.

RumovZ:
Sorry for the delay. I didn't get around to review this thoroughly, but you seem to have addressed the most important stuff. If you want to merge, I won't stand in the way.

randompersona1:
I think what I'll do then is add a "(Beta)" text to the groupbox. I'm sure there are *some* problems that I'll just solve when they pop up.

RumovZ (approved):

src/usdb_syncer/sync_meta.py
```diff
@@ -186,6 +188,8 @@ def try_from_file(cls, path: Path) -> SyncMeta | None:
                 pinned=bool(dct.get("pinned", False)),
                 txt=Resource.from_nested_dict(dct["txt"]),
                 audio=Resource.from_nested_dict(dct["audio"]),
+                vocals=Resource.from_nested_dict(dct["vocals"]),
+                instrumental=Resource.from_nested_dict(dct["instrumental"]),
```
RumovZ, line 192:
```suggestion
                vocals=Resource.from_nested_dict(dct.get("vocals")),
                instrumental=Resource.from_nested_dict(dct.get("instrumental")),
```

@randompersona1, this breaks reading any collections from before this change. Also, r-Q6wyggsTg.usdb must not be modified so we catch exactly this kind of breakage. I wonder how we can ensure that so we don't forget again in the future. 🤔

bohning: I just fixed this in #580.

PR #574, opened 2026-04-04
Add version tooltip for yt-dlp, deno and ffmpeg.

bohning:
image


randompersona1:
I like it! Note that the ffmpeg version could change, since we don't have control over PATH. But I don't see how we could possibly invalidate the cache, so whatever

randompersona1 (approved):

PR #580, opened 2026-05-14
Separation improvements

bohning:


randompersona1:
The icon now doesnt change based on theme. Was that on purpose?

Other than that, LGTM

bohning:
Argh, that was not on purpose - I thought it was only for the custom help icon for the groupbox.

bohning:
@RumovZ @randompersona1 I just updated the dependencies to include the latest yt-dlp 2026.6.9. Also I fixed an issue with song tags not showing in the song table, I think this has been broken for a long time. With this PR merged, I think we can finally release 0.22.0. Could one of you check and update the changelog?

RumovZ (approved):

PR #591, opened 2026-06-24
Add setting to limit background image size.

bohning:
image


src/usdb_syncer/resource_dl.py
```diff
@@ -521,10 +522,30 @@ def _adjust_contrast(image: Image.Image, meta_tags: ImageMetaTags) -> Image.Imag
     return image
 
 
+def _should_apply_max_size(
+    kind: ImageKind, max_width: CoverMaxSize | BackgroundMaxSize, image_width: int
```
RumovZ: This can be simplified by having a common base class like MaxSize that implements _should_apply_max_size().

bohning: Like this?

RumovZ (approved):

PR #594, opened 2026-07-01
Add a force redownload action.

bohning:
This action forces a redownload even of unchanged resources.


src/usdb_syncer/gui/resources/qt/download-force-white.svg
```diff
@@ -0,0 +1 @@
+<ns0:svg xmlns:ns0="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="white"><path d="M5,20l9,0l0,-2l-9,0m14,-9l-4,0l0,-6l-6,0l0,6l-4,0l7,7"/><path d="M18,13l2,0l0,5.5l-2,0l0,-5.5m0,9l0,-2l2,0l0,2l-2,0Z"/></ns0:svg>
```
RumovZ: Are these MD icons? GitHub can't render them.
Incidentally, I just noticed another icon that doesn't seem to be from the standard icon set. As you can see, it gets distorted so much that it's hardly recognisable. This is easily avoided by just using MD icons which always follow certain style rules.

image

RumovZ: Actually, I could be mistaken and it is an MD icon, just that it represents a letter rather than a cross.

bohning: Yes and no. MD just had a [download icon with a lock](https://pictogrammers.com/library/mdi/icon/download-lock/), so I manually exchanged the lock with a [thick exclamation mark](https://pictogrammers.com/library/mdi/icon/exclamation-thick/) to create the new icon for forced downloads. The issue was the custom namespace, it should now render properly in the diff.

bohning: > Actually, I could be mistaken and it is an MD icon, just that it represents a letter rather than a cross.

Yes, I think it is [this](https://pictogrammers.com/library/mdi/icon/alpha-x/) one.

RumovZ: Still doesn't render.

Could you replace the x icon with it with https://pictogrammers.com/library/mdi/icon/close/ or https://pictogrammers.com/library/mdi/icon/window-close/ (already checked in) while you're at it?

bohning: Strange, renders for me now:
image

RumovZ: Thanks!

I'm on a different device now and it works, so I assume it's fine.

RumovZ (approved):

